PR #21, opened 2017-10-16
Add ability to set Look & Feel via environment variable.

michaelsavich:
Hi, I saw your project and thought it looked interesting. I noticed you hardcoded the Look & Feel, so I added the ability to set the it using the environment variable COL_RECAF_LAF. Note that as far as I know this only seems to work when the app is launched from the command line.

### DIFF
--- a/src/me/coley/recaf/Recaf.java
+++ b/src/me/coley/recaf/Recaf.java
@@ -5,6 +5,7 @@ import java.io.File;
 import java.io.IOException;
 
 import javax.swing.UIManager;
+import javax.swing.UnsupportedLookAndFeelException;
 
 import org.objectweb.asm.tree.ClassNode;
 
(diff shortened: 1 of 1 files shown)


Col-E:
Thanks! How would this look when invoked from command line?

michaelsavich:
Here's an image of how it looks with the macOS "Aqua" look and feel: https://imgur.com/a/u4fRG

Also, setLookAndFeel is a method I added to Recaf.java. If you pass null to setLookAndFeel, it uses the default. I chose that approach to minimize the number of if-else branches.

Col-E:
Thanks for the contribution!

Col-E:
Expanded upon the idea in https://github.com/Col-E/Recaf/commit/98b6baebbd078ebdb723993f0ccae408f8b26650, which moves LAF application to the options menu. The themes are populated from UiManager's installed theme list. You can even swap LAF's at runtime too! 

For example I started with Nimbus and swapped to windows *(Both with different windows themes applied)*

javaw_2017-10-16_04-10-17
javaw_2017-10-16_04-13-05
javaw_2017-10-16_04-11-53

PR #35, opened 2018-01-26
fix pom.xml resource path

CreateRemoteThread:
1) pom.xml fix, to include en.json in the default build (i.e. "git clone, mvn package")
2) Error handling robustness for when getType() fails

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
 			<resource>
 				<directory>src</directory>
 				<includes>
-					<include>resources/*.png</include>
+					<include>resources/**/*</include>
 				</includes>
 			</resource>
 		</resources>
--- a/src/me/coley/recaf/ui/component/list/OpcodeCellRenderer.java
+++ b/src/me/coley/recaf/ui/component/list/OpcodeCellRenderer.java
@@ -69,16 +69,11 @@ public class OpcodeCellRenderer implements RenderFormatter<AbstractInsnNode>, Op
 		case AbstractInsnNode.TYPE_INSN:
 			// Add type name to string
 			TypeInsnNode insnType = (TypeInsnNode) ain;
-			Type typeDeclared = null;
-			String typeDeclaredStr = "";
-			try {
-				typeDeclared = Type.getType(insnType.desc);
-				typeDeclaredStr = getTypeStr(typeDeclared, options);
-			}
-			catch(Exception e) {
-				typeDeclaredStr = insnType.desc;
-			}
-			s += color(theme.opcodeTypeDefinition, italic(" " + typeDeclaredStr));
+			String typeDesc = insnType.desc;
+			if (options.opcodeSimplifyDescriptors && typeDesc.contains("/")) {
+				typeDesc = typeDesc.substring(typeDesc.lastIndexOf("/") + 1);
+			}	
+			s += color(theme.opcodeTypeDefinition, italic(" " + typeDesc));
 			break;
 		case AbstractInsnNode.FIELD_INSN:
 			FieldInsnNode insnField = (FieldInsnNode) ain;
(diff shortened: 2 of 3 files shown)


src/me/coley/recaf/ui/component/list/RenderFormatter.java
```diff
@@ -89,10 +89,15 @@ default String getTypeStr(Type type, ConfUI options) {
 		if (!s.contains("(") && (s.length() == 1 || s.startsWith("L") || s.startsWith("["))) {
 			s = type.getClassName();
 		}
+    if(s == null)
+    {
+      System.out.println("Type " + type + " has no descriptor or class");
```
Col-E: For future pulls using the logging impl will write to the log file and console at once. This could be `WARN` level *(Recaf.INSTANCE.logging.<level-for-method-name>(message)*. 
Original source also uses tabs instead of spaces, so for consistency keeping the scheme the same across all classes would be appreciated.

pom.xml
```diff
@@ -67,7 +67,7 @@
 			<resource>
 				<directory>src</directory>
 				<includes>
-					<include>resources/*.png</include>
+					<include>resources/**/*</include>
```
Col-E: Funny thing is that I never noticed this because it seems to add it anyways *(at least on my end)*. Nice catch.

src/me/coley/recaf/ui/component/list/OpcodeCellRenderer.java
```diff
@@ -69,7 +69,16 @@ public String getOpcodeText(OpcodeList list, AbstractInsnNode ain) {
...
+      Type typeDeclared = null;
+      String typeDeclaredStr = "";
+      try{
+        typeDeclared = Type.getType(insnType.desc);
+			  typeDeclaredStr = getTypeStr(typeDeclared, options);
+      }
+      catch(Exception e)
+      {
+        typeDeclaredStr = insnType.desc;
+      }
```
Col-E: No issues here, but curious to cases where this fails. Do you have any samples you can share that cause this? It may be solve-able without a try-catch if there's some sort of pattern to the failure.

CreateRemoteThread:
I've changed the tabs to be spaces. For a test case, take a look at https://drive.google.com/open?id=1XQvNM7ip4FkcW5018BRYPsRcMHkXfKeq (pw is your Github username). Trying to look at the opcodes of X.a(byte[] paramArrayOfByte, int paramInt) triggered this initially for me.

Col-E:
Ok so the issue is my fault.
```java
140: new #725 // class s
```

When I wrote this I must have been assuming `TypeInsnNode#desc` was formatted like a field descriptor, but it's actually just an internal name so the usage of `Type#getType(desc)` I have in that switch block is wrong. The proper block code should be:

```java
		case AbstractInsnNode.TYPE_INSN:
			// Add type name to string
			TypeInsnNode insnType = (TypeInsnNode) ain;
			String typeDesc = insnType.desc;
			if (options.opcodeSimplifyDescriptors && typeDesc.contains("/")) {
				typeDesc = typeDesc.substring(typeDesc.lastIndexOf("/") + 1);
			}	
			s += color(theme.opcodeTypeDefinition, italic(" " + typeDesc));
			break;
```


The NPE is from the usage of `type.getClassName` in `getTypeStr` because of the logic of `Type#getType` 
resulting in names such as `my/class/Name` being treated as method descriptors. For a `Type` with a `METHOD`  `sort` `getClassName` will return `null`. 
```java
default String getTypeStr(Type type, ConfUI options) {
	String s = type.getDescriptor();
	// Check if field type. If so, then format as class name.
	if (!s.contains("(") && (s.length() == 1 || s.startsWith("L") || s.startsWith("["))) {
		s = type.getClassName();
	}
	// ... rest of the method
```
```java
public String getClassName() {
    switch (sort) {
    // primitive and array cases
    case OBJECT: // what we want, but not called since sort is METHOD
        return new String(buf, off, len).replace('/', '.');
    default:
        return null;
    }
}
```
```java
private static Type getType(final char[] buf, final int off) {
    int len;
    // buff[0] is 's' which does not match any case here, 
    // resulting in returning a constructed `METHOD` type with an obviously non-method value in `buf`
    switch (buf[off]) {
    // cases for primitives, arrays, etc.
    default:
        return new Type(METHOD, buf, off, buf.length - off);
    }
}
```

You can make the fix part of the pull if you wish, I'm off to bed.

Col-E:
Thanks!

PR #39, opened 2018-02-02
Fix a encoding problem

andylizi:
`Charset.defaultCharset()` will return different charsets on different platforms. For example, this method returns `GBK` in China, which will result in some unintelligible texts like this:
Recaf with Chinese language file

I hope my expression are understandable, I'm not native speakers.


Col-E:
Thanks!

PR #85, opened 2018-09-27
Add Simplified-Chinese translation

andylizi:


andylizi:
Do you think I should translate the ASM config part? There're a lot of terms there I'm not sure how to express in Chinese/should not translate into Chinese. If I only translate some of them it'll look pretty weird.

By the way, I found some unused translation key. Not sure if I translated them correctly because I don't know where they were being used.
```
update.fail.nopom.dev
ui.edit.method.insert
ui.history.index
ui.history.time
ui.history.stacksize
ui.info.other
```

Col-E:
Those items were removed recently when the history window got updated.

Current: 
[screenshots/save-state-view.png](https://col-e.github.io/Recaf/screenshots/save-state-view.png)

Col-E:
`update.fail.nopom.dev` - This was used in an older iteration of the auto-updater. I'll delete it from the translation files.

`ui.edit.method.insert` - Looking at this is kinda confusing. The name would imply it should relate to methods but instead if you look at the sub-items such as `ui.edit.method.insert.title` its used for opcodes. This is probably an oversight on my part. `ui.edit.method.insert` is unused.

`ui.info.other` is unused too. I was going to have some extra information on a separate tab next to the logging tab, but never got around to actually doing anything with it. If you have any ideas on what can be put there I'm open to suggestions.

andylizi:
A small typo in [en.json#L193](../blob/40c8b0736098a255a749fdafc350ca6bb0fb78de/src/resources/lang/en.json#L193):
```json
"ui.bean.class.outermethod.desc": "Name of the method in the outer class that this class was written in. This is class is anonymous if non-null."
```
`This is class is anonymous if non-null.` ->
`This class is anonymous if non-null.`

Col-E:
If you find any issues do add them to the PR.

andylizi:
I don't want to change any file besides `zh-cn.json` because they may cause conflicts when merging. This is WIP so it'll be some time before all commits can be merged, and by then there may be... you know

(ah such pain expressing myself in foreign language..)

Col-E:
For `"ui.bean.class.outermethod.desc"` I re-read it and replaced it with the exact wording of the ClassNode documentation for the field.

andylizi:
All translation were basically completed.

I don't think I have enough energy and patience to handle the CFR part, and, it's really not *that* important, sooooo I'll just go and pretend it's all done :smirk:

Col-E:
:+1: merging, awesome work!

PR #90, opened 2018-09-28
 Refactor access util & AccessButton

andylizi:
### Screenshots
class-access
innerclass-access
field-access
method-access

I tried to write some documents, but it's really hard for a non-native English speaker. Eventually I had to give up on that :(


Col-E:
:+1:

PR #93, opened 2018-09-30
 Make enums translatable

andylizi:
### Screenshots
search-string-mode
class-version
type-annotation-reftype
config-update


andylizi:
By the way, your Github account uses `Matt <Col-E@users.noreply.github.com>` for web-based git operations (e.g. auto-generated merge commit when merging PR) but your git CLI uses `Col-E <mcoley2@gmu.edu>` when committing, so git thinks there're two person. You may want to correct this.

PR #107, opened 2018-11-21
Complete support for all allowed annotation value types

andylizi:
#### Screenshot
annotations

#### Sample code
```java
@Target({ElementType.TYPE})
@Retention(RetentionPolicy.RUNTIME)
@TestAnnotation(values1 = {"v1", "v2"}, values2 = {Exception.class, Error.class}, values4 = @Retention(RetentionPolicy.CLASS))
@TestAnnotation2(@TestAnnotation(values2 = {}, values3 = {@Target(ElementType.TYPE)}))
public @interface TestAnnotation {
    String[] values1() default {};

    Class<?>[] values2();

    Target[] values3() default {};

    Retention values4() default @Retention(RetentionPolicy.SOURCE);
}

@Target({ElementType.TYPE, ElementType.ANNOTATION_TYPE})
@Retention(RetentionPolicy.RUNTIME)
public @interface TestAnnotation2 {
    TestAnnotation value();
}
```

### DIFF
--- a/src/me/coley/recaf/ui/FormatFactory.java
+++ b/src/me/coley/recaf/ui/FormatFactory.java
@@ -5,6 +5,7 @@ import org.objectweb.asm.Type;
 import org.objectweb.asm.tree.*;
 import static org.objectweb.asm.Opcodes.*;
 
+import java.lang.annotation.Annotation;
 import java.util.List;
 import java.util.Set;
 
@@ -111,6 +112,11 @@ public class FormatFactory {
 	 */
 	public static TextHBox annotation(AnnotationNode item) {
 		TextHBox t = new TextHBox();
+		annotation(t, item);
+		return t;
+	}
+
+	private static void annotation(TextHBox t, AnnotationNode item) {
 		if (item.desc != null) {
 			addRaw(t, "@");
 			addType(t, Type.getType(item.desc));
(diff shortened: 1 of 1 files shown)


andylizi:
By the way, do you think we should omit the default name "value" with something like
```diff
diff --git a/src/me/coley/recaf/ui/FormatFactory.java b/src/me/coley/recaf/ui/FormatFactory.java
index 6fd83ab..9e607d1 100644
--- a/src/me/coley/recaf/ui/FormatFactory.java
+++ b/src/me/coley/recaf/ui/FormatFactory.java
@@ -136,6 +136,8 @@ public class FormatFactory {
                                String name = (String) item.values.get(i);
                                Object value = item.values.get(i + 1);
-                               addName(t, name);
-                               addRaw(t, "=");
+                               if (!"value".equals(name)) {
+                                       addName(t, name);
+                                       addRaw(t, "=");
+                               }
                                if (value instanceof String) {
                                        addString(t, value.toString());
```
I can see both pros and cons of it. It's hard to decide.

Col-E:
```java
// @anno with multiple fields
public @interface Test1 {
	String[] value();
	String[] a() default {};
}
// Source visualization - sets value of "value"
@Test1({""})
// Recaf visualization - with included "value="
@Test1(value=[String...],a=[String...],)
```
I'm personally for keeping it in situations like this. However if there is only one field in the annotation then removing it makes sense:
```java
// @anno with a single field
public @interface Test2 {
	String[] value();
}
// Source visualization - sets value of "value"
@Test2({""})
// Recaf visualization - without included "value="
@Test2([String...])
```

andylizi:
Defaulted value won't appear in the bytecode after compilation, so the visualization will actually look like this:
```java
public @interface Test1 {
	String[] value();
	String[] a() default {};
}
// Source visualization
@Test1({""})
// Recaf visualization
@Test1(value=[String...])
```

---

#### Patch
```diff
diff --git a/src/me/coley/recaf/ui/FormatFactory.java b/src/me/coley/recaf/ui/FormatFactory.java
index 6fd83ab..ec2b275 100644
--- a/src/me/coley/recaf/ui/FormatFactory.java
+++ b/src/me/coley/recaf/ui/FormatFactory.java
@@ -136,6 +136,8 @@ public class FormatFactory {
                                String name = (String) item.values.get(i);
                                Object value = item.values.get(i + 1);
-                               addName(t, name);
-                               addRaw(t, "=");
+                               if (max > 2 || !"value".equals(name)) {
+                                       addName(t, name);
+                                       addRaw(t, "=");
+                               }
                                if (value instanceof String) {
                                        addString(t, value.toString());
```

#### Sample code
```java
public @interface Test1 {
	@Test1("abc")
	String value();

	@Test1(value = "abc", foo = "bar")
	String foo() default "default value of foo()";
}

public @interface Test2 {
	@Test2(foo = "bar")
	String value() default "default value of value()";

	@Test2(value = "abc", foo = "bar")
	String foo();
}
```

#### Result
It'll look exactly the same in source and in Recaf, expect the quotes. (Why does string literal value doesn't have quotes wrapped around them anyway?

Col-E:
If the string literal is missing quotes that is an oversight. What you just posted is exactly what I was suggesting the result to look like.

Issue #134, opened 2019-06-16
Auto-completion doesn't suggest any of the bootstrap classes


andylizi:
I discovered this during some refactoring.
https://github.com/Col-E/Recaf/blob/f0c541b00176de915c3e1514ee03f0c1c16431c7/src/main/java/me/coley/recaf/util/Classpath.java#L129-L174

I had found a way to solve it under Oracle JVM 8, but it won't work in Java 9, or in other JVM implementations. So that's ANOTHER compatibility problem😓. Maybe there's another solution that I've missed?

Col-E:
I was hoping Guava's `ClassPath` would work, but I guess we're not allowed to have nice things.

andylizi:
I'm thinking about giving up on auto-scanning and just ask nicely for library locations, like what ProGuard did:

![](https://user-images.githubusercontent.com/12008103/62463155-da417100-b7bb-11e9-82c0-00805a581371.png)

It would save a lot of headaches, and had the benefit of choosing other versions of JDK and adding other library jars. Currently Recaf can only open one jar at a time, so Assembler can't suggest classes from any of the other dependencies.

We can still place the automatically acquired bootstrap classpath in the list of libraries if the list is empty and JRE version is below 9, like a default setting (which I think is also what ProGuard did).

Col-E:
> We can still place the automatically acquired bootstrap classpath in the list of libraries if the list is empty and JRE version is below 9, like a default setting (which I think is also what ProGuard did).

:+1: 

> Currently Recaf can only open one jar at a time

That was one of the first things I wanted to change in `2.0.0`, it'll make this super trivial in the future.

Col-E:
Fixed in #162 

For maintainability purposes should we still consider the proguard approach for later?

xxDark:
Closing since its been fixed/implemented a long time ago, pending support for auto-completion is pending on https://github.com/Col-E/Recaf/pull/564

PR #162, opened 2019-08-04
Scan boot classpath on JVM 9 and later

xxDark:
## What's new

Returns scanning of boot classpath on JVM 9 and later versions, checking of version is done by checking `java.class.version` system property

### DIFF
--- a/src/main/java/me/coley/recaf/util/Classpath.java
+++ b/src/main/java/me/coley/recaf/util/Classpath.java
@@ -137,7 +137,7 @@ public class Classpath {
 			// auto-completion will not able to suggest internal names of any of the classes under java.*,
 			// which will largely reduce the effectiveness of the feature.
 			if (!checkBootstrapClass()) {
-				int vmVersion = Integer.parseInt(System.getProperty("java.class.version")) - 44;
+				float vmVersion = Float.parseFloat(System.getProperty("java.class.version")) - 44;
 				if (vmVersion < 9) {
 					try {
 						Method method = ClassLoader.class.getDeclaredMethod("getBootstrapClassPath");
@@ -161,6 +161,7 @@ public class Classpath {
 						Object bootstrapClasspath = field.get(bootLoader);
 						scanBootstrapClasspath(field, classLoader, bootstrapClasspath);
 					} catch (ReflectiveOperationException | SecurityException e) {
+						e.printStackTrace();
 						throw new ExceptionInInitializerError(e);
 					}
 				}
--- a/src/main/java/me/coley/recaf/util/Classpath.java
+++ b/src/main/java/me/coley/recaf/util/Classpath.java
@@ -16,6 +16,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -198,9 +199,14 @@ public class Classpath {
 								export.invoke(impl, name);
 							}
 						}
-						MethodHandle mapHandle = lookup.findStaticGetter(Class.forName("jdk.internal.reflect.Reflection"), "fieldFilterMap", Map.class);
-						Map<Class<?>, Set<String>> fieldFilterMap = (Map<Class<?>, Set<String>>) mapHandle.invokeExact();
+
+						Class<?> reflectionClass = Class.forName("jdk.internal.reflect.Reflection");
+						MethodHandle getMapHandle = lookup.findStaticGetter(reflectionClass, "fieldFilterMap", Map.class);
+						// Map is immutable, thanks Oracle
+						Map<Class<?>, Set<String>> fieldFilterMap = new HashMap<>((Map<Class<?>, Set<String>>) getMapHandle.invokeExact());
 						fieldFilterMap.remove(Field.class);
+						lookup.findStaticSetter(reflectionClass, "fieldFilterMap", Map.class)
+								.invokeExact(fieldFilterMap);
 
 						Method method = Class.forName("jdk.internal.loader.ClassLoaders").getDeclaredMethod("bootLoader");
 						method.setAccessible(true);
--- a/src/main/java/me/coley/recaf/util/Classpath.java
+++ b/src/main/java/me/coley/recaf/util/Classpath.java
@@ -9,6 +9,7 @@ import java.lang.invoke.MethodHandles;
 import java.lang.invoke.MethodType;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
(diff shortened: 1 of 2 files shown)


Col-E:
Seems to not be working on Java 12. Does this work on Java 9?

```
Caused by: java.lang.IllegalArgumentException: Can not set final jdk.internal.loader.URLClassPath field jdk.internal.loader.BuiltinClassLoader.ucp to java.net.URLClassLoader
	... INTERNAL CALLS ...
	at java.base/java.lang.reflect.Field.set(Field.java:780)
	at me.coley.recaf.util.Classpath$ClassPathScanner.scanBootstrapClasspath(Classpath.java:235)
	at me.coley.recaf.util.Classpath$ClassPathScanner.scan(Classpath.java:206)
```

I noticed that this sets the field `URLClassPath ucp` to a `URLClassLoader`, but this fails because the types are incompatible.

xxDark:
Should be fixed now, can't test right now.

xxDark:
Also fixes #134

Col-E:
Debugging on Java 12 currently, this does correctly scan the boot classpath on JDK 9+ (:+1: ) but I can't seem to verify that #134 is fixed. 

When running `mvn test` the `AutoCompleteTest` class is still skipped because `Classpath.isBootstrapClassesFound()` still returns false. 

image

The evaluation output should have names like `java.lang.Object`

Col-E:
I can merge now if we want to tackle #134 later in a separate PR

andylizi:
> Debugging on Java 12 currently, this does correctly scan the boot classpath on JDK 9+

Huh? When I tried it on Java 11, it didn't work because `bootstrapClasspath` is `null` (like I said above).

Did Java 12 change something? I only have 11 and 1.8 installed because they're the LTS versions.

Col-E:
> Did Java 12 change something?

I was mistaken, see the edit.

xxDark:
I guess because of Jigsaw module system boot classpath does not contain `java.*` and all other packages anymore, will look on what I can do with that

xxDark:
Got it

Col-E:
image
:ok_hand:

src/main/java/me/coley/recaf/util/Classpath.java
```diff
@@ -138,38 +146,112 @@ public void scan(ClassLoader classLoader) {
...
+								.findVirtual(moduleClass, "getPackages", MethodType.methodType(Set.class));
+						MethodHandle getModule = lookup
+								.findVirtual(Class.class, "getModule", MethodType.methodType(moduleClass));
+						MethodHandle getLayer = lookup
+								.findVirtual(moduleClass, "getLayer", MethodType.methodType(layer));
+						MethodHandle layerModules = lookup
+								.findVirtual(layer, "modules", MethodType.methodType(Set.class));
+						MethodHandle unnamedModule = lookup
+								.findVirtual(ClassLoader.class, "getUnnamedModule", MethodType.methodType(moduleClass));
+						Set modules = new HashSet();
```
andylizi:
```suggestion
						Set<Object> modules = new HashSet<>();
```

andylizi:
```suggestion
						Set<Object> modules = new HashSet<>();
```

src/main/java/me/coley/recaf/util/Classpath.java
```diff
@@ -138,38 +146,112 @@ public void scan(ClassLoader classLoader) {
...
+
+						Object ourModule = getModule.invoke(Classpath.class);
+						Object ourLayer = getLayer.invoke(ourModule);
+						if (ourLayer != null) {
+							modules.addAll((Set) layerModules.invoke(ourLayer));
+						}
+						modules.addAll(
+								(Set) layerModules
```
andylizi:
```suggestion
								(Collection<?>) layerModules
```

src/main/java/me/coley/recaf/util/Classpath.java
```diff
@@ -138,38 +146,112 @@ public void scan(ClassLoader classLoader) {
...
+						for (Object impl : modules) {
+							for (String name : (Set<String>) getPackages.invoke(impl)) {
+								export.invoke(impl, name);
+							}
+						}
+
+						Class<?> reflectionClass = Class.forName("jdk.internal.reflect.Reflection");
+						MethodHandle getMapHandle = lookup.findStaticGetter(reflectionClass, "fieldFilterMap", Map.class);
+						// Map is immutable, thanks Oracle
+						Map<Class<?>, Set<String>> fieldFilterMap = new HashMap<>((Map<Class<?>, Set<String>>) getMapHandle.invokeExact());
```
andylizi:
```suggestion
						Map<Class<?>, ?> fieldFilterMap = new HashMap<>((Map<Class<?>, ?>) getMapHandle.invokeExact());
```

They changed some code in `Reflection` in JDK 12: [before](https://github.com/openjdk/jdk11u/blob/71f32ecc0fa1d0c552c40b1222412a70678dea19/src/java.base/share/classes/jdk/internal/reflect/Reflection.java#L46-L58) and  [after](https://github.com/openjdk/jdk12u/blob/fca351466233b0ac5867f3cd690897df3d707726/src/java.base/share/classes/jdk/internal/reflect/Reflection.java#L45-L63), so 

1. The generic signature of `fieldFilterMap` should be `Map<Class<?>, ?>` for compatibility between `Map<Class<?>, String[]>` and `Map<Class<?>, Set<String>>` (we are only using the key anyway).
2. What's more, it's altogether unnecessary to do this before Java 12 - older versions didn't filter out `Field` access.

src/main/java/me/coley/recaf/util/Classpath.java
```diff
@@ -138,38 +146,112 @@ public void scan(ClassLoader classLoader) {
...
+						listReader.setAccessible(true);
+						Field bootModulesField = bootLoader.getClass().getSuperclass().getDeclaredField("packageToModule");
+						bootModulesField.setAccessible(true);
+						Collection<?> packageToModule = ((Map<String, ?>) bootModulesField.get(null)).values();
+						for (Object module : packageToModule) {
+							// Scan it!
+							Object ref = mref.invoke(module);
+							Object reader = openReader.invoke(ref);
+							Stream<String> list = (Stream<String>) listReader.invoke(reader);
+							list.parallel().filter(s -> s.endsWith(".class"))
```
andylizi:
```suggestion
							list.filter(s -> s.endsWith(".class"))
```

[Should I always use a parallel stream when possible? - Stack Overflow](https://stackoverflow.com/a/20375622/6238618)

Col-E:
Any more commits to add or are we all ready to merge?

xxDark:
> 
> 
> Any more commits to add or are we all ready to merge?

Ready

Col-E:
Thank  for the PR :+1:

PR #164, opened 2019-08-09
Code cleanup

xxDark:
## What's new
Code cleanup, see commits for changes

### DIFF
--- a/src/main/java/me/coley/recaf/parse/source/SourceCode.java
+++ b/src/main/java/me/coley/recaf/parse/source/SourceCode.java
@@ -1,4 +1,4 @@
-package me.coley.recaf.workspace;
+package me.coley.recaf.parse.source;
 
 import com.github.javaparser.*;
 import com.github.javaparser.ast.CompilationUnit;
@@ -7,6 +7,8 @@ import com.github.javaparser.ast.body.TypeDeclaration;
 import com.github.javaparser.ast.expr.*;
 import com.github.javaparser.ast.nodeTypes.NodeWithName;
 import me.coley.recaf.util.StringUtil;
+import me.coley.recaf.workspace.JavaResource;
+import me.coley.recaf.workspace.Workspace;
 
 import java.util.*;
 import java.util.stream.Collectors;
--- a/src/main/java/me/coley/recaf/parse/source/SourceCode.java
+++ b/src/main/java/me/coley/recaf/parse/source/SourceCode.java
@@ -136,7 +136,7 @@ public class SourceCode {
 		if (packageName != null)
 			return packageName;
 		// fetch package
-		return unit.getPackageDeclaration().map(NodeWithName::getNameAsString).orElse(DEFAULT_PACKAGE);
+		return packageName = unit.getPackageDeclaration().map(NodeWithName::getNameAsString).orElse(DEFAULT_PACKAGE);
 	}
 
 	/**
--- a/src/main/java/me/coley/recaf/search/SearchCollector.java
+++ b/src/main/java/me/coley/recaf/search/SearchCollector.java
@@ -66,7 +66,7 @@ public class SearchCollector {
 		// Get results of multiple queries that have either the same parent context
 		// (or just same context in some cases)
 		// TODO: remove asMap() call?
-		return (List<SearchResult>) results.asMap().values().stream()
+		return new ArrayList<>(results.asMap().values().stream()
 				.reduce((a, b) -> {
 					// Set so we don't get duplicates
 					Set<SearchResult> ret = new LinkedHashSet<>();
@@ -96,14 +96,14 @@ public class SearchCollector {
 					}
 					// Back to list
 					return new ArrayList<>(ret);
-				}).get();
+				}).get());
 	}
 
 	/**
 	 * @return Flattened list of the {@link #getResultsMap() result map}.
 	 */
 	public List<SearchResult> getAllResults() {
-		return (List<SearchResult>) results.values();
+		return new ArrayList<>(results.values());
 	}
 
 	/**
--- a/src/main/java/me/coley/recaf/search/SearchCollector.java
+++ b/src/main/java/me/coley/recaf/search/SearchCollector.java
@@ -1,14 +1,14 @@
 package me.coley.recaf.search;
 
+import com.google.common.collect.ListMultimap;
+import com.google.common.collect.MultimapBuilder;
 import me.coley.recaf.util.ClassUtil;
 import me.coley.recaf.workspace.Workspace;
 import org.objectweb.asm.*;
 import org.objectweb.asm.tree.*;
 
 import java.util.*;
-import java.util.function.Function;
 import java.util.function.Supplier;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import static org.objectweb.asm.ClassReader.*;
@@ -32,7 +32,8 @@ import static org.objectweb.asm.ClassReader.*;
  */
 public class SearchCollector {
 	public static final int ACC_NOT_FOUND = -1;
-	private final Map<Query, List<SearchResult>> results = new LinkedHashMap<>();
+	private final ListMultimap<Query, SearchResult> results = MultimapBuilder
+			.linkedHashKeys(2).arrayListValues().build();
 	private final Workspace workspace;
 	private final Collection<Query> queries;
 
@@ -52,7 +53,7 @@ public class SearchCollector {
 	/**
 	 * @return Map of queries to their results.
 	 */
-	public Map<Query, List<SearchResult>> getResultsMap() {
+	public ListMultimap<Query, SearchResult> getResultsMap() {
 		return results;
 	}
 
@@ -64,7 +65,8 @@ public class SearchCollector {
 	public List<SearchResult> getOverlappingResults() {
 		// Get results of multiple queries that have either the same parent context
 		// (or just same context in some cases)
-		return results.values().stream()
+		// TODO: remove asMap() call?
+		return (List<SearchResult>) results.asMap().values().stream()
 				.reduce((a, b) -> {
 					// Set so we don't get duplicates
 					Set<SearchResult> ret = new LinkedHashSet<>();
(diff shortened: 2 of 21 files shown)


src/main/java/me/coley/recaf/workspace/History.java
```diff
@@ -63,7 +63,7 @@ public void clear() {
 	 * @return Array of timestamps of each tracked change.
 	 */
 	public Instant[] getFileTimes() {
-		return times.toArray(new Instant[0]);
+		return times.toArray(new Instant[times.size()]);
```
andylizi: Technically, `toArray(new T[0])` is faster...

[java - .toArray(new MyClass[0]) or .toArray(new MyClass[myList.size()])? - Stack Overflow](https://stackoverflow.com/a/29444594/6238618)

Also according to [Intellij IDEA Inspection `ToArrayCallWithZeroLengthArrayArgument`](https://github.com/JetBrains/intellij-community/blob/master/plugins/InspectionGadgets/src/inspectionDescriptions/ToArrayCallWithZeroLengthArrayArgument.html):

> In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow. **However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version.** Also passing pre-sized array is **dangerous for a concurrent or synchronized collection** as a data race is possible between the `size` and `toArray` call which may result in extra nulls at the end of the array, if the collection was concurrently shrunk during the operation.
> 
> This inspection allows to follow the uniform style: either **using an empty array (which is recommended in modern Java)** or using a pre-sized array (which might be faster in older Java versions or non-HotSpot based JVMs).

src/main/java/me/coley/recaf/search/SearchCollector.java
```diff
@@ -133,7 +132,7 @@ void addMatched(Context<?> context, Query quert) {
 		List<SearchResult> matched = quert.getMatched();
 		if(context != null)
 			matched.forEach(res -> res.setContext(context));
-		results.computeIfAbsent(quert, p -> new ArrayList<>()).addAll(matched);
+		results.computeIfAbsent(quert, p -> new ArrayList<>(matched.size())).addAll(matched);
```
andylizi: Why don't we take it a step further and just use [`Multimap`](https://guava.dev/releases/28.0-jre/api/docs/com/google/common/collect/Multimap.html) from Guava instead :wink:

```java
ListMultimap<Query, SearchResult> results = MultimapBuilder
		.linkedHashKeys(2).arrayListValues().build();
```

Oh, and there's a typo in the parameter `quert` by the way.

src/main/java/me/coley/recaf/parse/source/SourceCode.java
```diff
@@ -135,9 +136,7 @@ public String getPackage() {
 		if (packageName != null)
 			return packageName;
 		// fetch package
-		if(unit.getPackageDeclaration().isPresent())
-			return packageName = unit.getPackageDeclaration().get().getNameAsString();
-		return packageName = DEFAULT_PACKAGE;
+		return unit.getPackageDeclaration().map(NodeWithName::getNameAsString).orElse(DEFAULT_PACKAGE);
```
andylizi:
```suggestion
		return packageName = unit.getPackageDeclaration().map(NodeWithName::getNameAsString).orElse(DEFAULT_PACKAGE);
```

src/main/java/me/coley/recaf/workspace/History.java
xxDark: > 
> 
> Technically, `toArray(new T[0])` is faster...
> 
> [java - .toArray(new MyClass[0]) or .toArray(new MyClass[myList.size()])? - Stack Overflow](https://stackoverflow.com/a/29444594/6238618)
> 
> Also according to [Intellij IDEA Inspection `ToArrayCallWithZeroLengthArrayArgument`](https://github.com/JetBrains/intellij-community/blob/master/plugins/InspectionGadgets/src/inspectionDescriptions/ToArrayCallWithZeroLengthArrayArgument.html):
> 
> > In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow. **However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version.** Also passing pre-sized array is **dangerous for a concurrent or synchronized collection** as a data race is possible between the `size` and `toArray` call which may result in extra nulls at the end of the array, if the collection was concurrently shrunk during the operation.
> > This inspection allows to follow the uniform style: either **using an empty array (which is recommended in modern Java)** or using a pre-sized array (which might be faster in older Java versions or non-HotSpot based JVMs).

In this case, it's not.
Look at `ArrayList#toArray(T[])` implementation

src/main/java/me/coley/recaf/util/ClassUtil.java
```diff
@@ -124,6 +128,6 @@ public FieldVisitor visitField(int access, String vname, String vdesc, String si
 	 * @return Standard class name.
 	 */
 	private static String normalize(String internal) {
```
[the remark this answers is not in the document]
xxDark: > 
> 
> It's both `private` and unused, so can be removed entirely.

Did not remove because it will may be used later, can mark as `@Deprecated`

src/main/java/me/coley/recaf/workspace/History.java
andylizi: > In this case, it's not.

But the example in the reference uses `ArrayList` too, why are 'this' case different?

Anyway. Yes, if we only consider the literal Java implementation it does seem faster. And that's why it's so counterintuitive. As articles said, there're other factors at work:

> ... since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version.

> To summarize: the JVM and JIT compiler contains several optimizations that enable it to cheaply create and initialize a new correctly sized array, and those optimizations can not be used if you create the array yourself.

> Bottom line: `toArray(new T[0])` seems faster, safer, and contractually cleaner, and therefore should be the default choice now. --- from [Conclusion, *Arrays of Wisdom of the Ancients*](https://shipilev.net/blog/2016/arrays-wisdom-ancients/#_conclusion)

Even if we disregard the analysis and reasoning, the benchmark result doesn't lie.

src/main/java/me/coley/recaf/util/ClassUtil.java
```diff
@@ -22,9 +22,14 @@
 	 */
 	public static ClassReader fromRuntime(String name) {
 		try {
-			Class<?> loaded = ClasspathUtil.getSystemClass(normalize(name));
-			return new ClassReader(loaded.getName());
-		} catch(ClassNotFoundException | IOException e) {
+			URL url = ClassLoader.getSystemResource(name.replace('.', '/').concat(".class"));
```
Col-E: I was under the assumption that using `concat` for optimization was no longer effective - https://stackoverflow.com/a/46485284

src/main/java/me/coley/recaf/search/SearchCollector.java
```diff
@@ -62,7 +65,8 @@ public SearchCollector(Workspace workspace, Collection<Query> queries) {
 	public List<SearchResult> getOverlappingResults() {
 		// Get results of multiple queries that have either the same parent context
 		// (or just same context in some cases)
-		return results.values().stream()
+		// TODO: remove asMap() call?
+		return (List<SearchResult>) results.asMap().values().stream()
```
Col-E: Casting like this is incorrect:

See results of `mvn test` _(`mvn test -Dcheckstyle.skip=true` in this case, see comments on `ClasspathUtil`)_

image

src/main/java/me/coley/recaf/search/SearchCollector.java
```diff
@@ -99,12 +103,7 @@ public SearchCollector(Workspace workspace, Collection<Query> queries) {
 	 * @return Flattened list of the {@link #getResultsMap() result map}.
 	 */
 	public List<SearchResult> getAllResults() {
-		if (results.isEmpty())
-			return Collections.emptyList();
-		return results.values().stream().reduce((a, b) -> {
-			a.addAll(b);
-			return a;
-		}).get();
+		return (List<SearchResult>) results.values();
```
Col-E: See prior casting comment

src/main/java/me/coley/recaf/util/ClassUtil.java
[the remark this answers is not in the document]
Col-E: If its usage has been inlined then its fine to remove it

src/main/java/me/coley/recaf/util/ClasspathUtil.java
```diff
@@ -150,38 +155,114 @@ public void scan(ClassLoader classLoader) {
...
+					try {
+
+						// Before we will do that, break into Jigsaw module system to grant full access
+						Class<?> moduleClass = Class.forName("java.lang.Module");
+						Class<?> layer = Class.forName("java.lang.ModuleLayer");
+						Field lookupField = MethodHandles.Lookup.class.getDeclaredField("IMPL_LOOKUP");
+						lookupField.setAccessible(true);
+						MethodHandles.Lookup lookup = (MethodHandles.Lookup) lookupField.get(null);
+						MethodHandle export = lookup
+								.findVirtual(moduleClass, "implAddOpens", MethodType.methodType(Void.TYPE, String.class));
```
Col-E: Lines like this surpass the Checkstyle limit of 120 chars.

See results of `mvn compile`

image

Col-E: Probably could import static methods to fix most of them.

xxDark:
I looked on `ListMultimap`, and I see no `clear` way to solve this runtime errors, maybe we should stick with `Map<Query, List<SearchResult>>` for now?

Col-E:
> 
> 
> I looked on `ListMultimap`, and I see no `clear` way to solve this runtime errors, maybe we should stick with `Map<Query, List<SearchResult>>` for now?

Since `ListMultimap` returns a `Collection<SearchResult>` that can be wrapped like so:

```java
	public List<SearchResult> getAllResults() {
		return new ArrayList<>(results.values());
	}
```

Same goes for `getOverlappingResults()`, just wrap the `get()` in an `new ArrayList<>(...)`. [This keeps the behavior the same.](https://user-images.githubusercontent.com/21371686/62823621-dd908000-bb60-11e9-8a2b-06f9da818f99.png)

Col-E:
Thanks! I'll merge if you don't have anything else you want to add to the PR.

xxDark:
> 
> 
> Thanks! I'll merge if you don't have anything else you want to add to the PR.

lgtm

PR #165, opened 2019-08-11
 Remove the redundant copying in overlapping results calculation

andylizi:
This is related to #164.

If there was [an `Lists.concat` method in Guava](//github.com/google/guava/pull/1029), we can avoid the copying in `getAllResults()` too, but I guess not...


Col-E:
Looks good :ok_hand:

andylizi:
> we could always make a `CollectionsUtil` class to make our own `Lists.concat`.

Yeah I considered it, but if here's the only place that uses it, it doesn't seem worth it. Its data size and calling frequency are not very significant anyway.

Issue #154, opened 2019-07-04
Runtime simulation


Col-E:
# Runtime simulation

The ability to simulate running code in a sandbox so that unknown classes and methods can be invoked without running potentially unsafe code _(Like deleting files)_. For primitives, strings, and other safe types this would allow users to watch / step through the execution and watch the changes applied to classes _(fields)_ and the method stack.  

## Use Cases

* Understanding the behavior of unknown logic without executing potentially malicious code
* Fetching values from simple methods
    * In combination with #151 could be used to automatically deobfuscate strings & other obfuscator patterns

## Implementation Outline:

**VirtualObj<T>**
- `T value`

**VirtualClass**
- `Map<Identifier, VirtualField> fields`
- `Map<Identifier, VirtualMethod> methods`

**VirtualMember**
- `identity()`
    - _abstract_, implemented by children. Used as the lookup key for **VirtualField** and **VirtualMethod**.

**VirtualField** _extends_ **VirtualMember**
- `VirtualClass value`

**VirtualMethod** _extends_ **VirtualMember**
- `Simulation getSimulation(VirtualObj... args)`
    - Generate **Simulation** object, does not run immediately

**Simulation**
- `List<VirtualInsn\> instructions`
- `List<VirtualObj\> vars`
- `Stack<VirtualObj\> stack`

**VirtualInsn** _models behavior from_ **AbstractInsnNode**
- `apply(Simulation)`
    - Applies modification to vars/stack as needed

***

Current thoughts on how the simulation API will be handled. There will probably be a lot of `VirtualClass` implementations for common core Java classes. At some point enough base classes should be implemented to have things referencing them work by auto-generated logic. For example, once `String` is implemented something like `StringUtils` should be able to have an auto-generated implementation work since all the outbound reference classes have a `VirtualClass` implementation. 

With this lookup strategy, loading loops would have to be counted somehow. Perhaps a combination of lazy-loading and keeping a global cache for library classes _(anything not residing in the primary Recaf input)_ would be worthwhile.

Also to reduce the complexity of certain simulations, being able to assign dummy values to things like field getters + method invokes may be a useful feature too.

Col-E:
## Method Simulation Outline:

**VirtualObj<T>**
- `T value`

**VirtualClass**
- `Map<Identifier, VirtualField> field`s
- `Map<Identifier, VirtualMethod> methods`

**VirtualMember**
- `identity()`
    - _abstract_, implemented by children

[...]

PR #169, opened 2019-09-13
[WIP] Initial concept of runtime simulation

xxDark:
## What's new
Initial concept for runtime simulation, as a goal for 2.0 #154 
I'm still not sure if it should be implemented that way, so all criticism/suggestions are welcome

### DIFF
--- a/src/main/java/me/xdark/recaf/runtime/Java9ClassDefiner.java
+++ b/src/main/java/me/xdark/recaf/runtime/Java9ClassDefiner.java
@@ -34,7 +34,7 @@ final class Java9ClassDefiner implements ClassDefiner {
             Class<?> unsafeClass = Class.forName("jdk.internal.misc.Unsafe", true, null);
             MH_DEFINE_CLASS = lookup.findVirtual(unsafeClass, "defineAnonymousClass",
                     MethodType.methodType(Class.class, Class.class, byte[].class, Object[].class));
-            JDK_UNSAFE = lookup.findGetter(unsafeClass, "theUnsafe", unsafeClass).invoke();
+            JDK_UNSAFE = lookup.findStatic(unsafeClass, "getUnsafe", MethodType.methodType(unsafeClass)).invoke();
         } catch (Throwable t) {
             throw new ExceptionInInitializerError(t);
         }
--- a/src/main/java/me/xdark/recaf/runtime/Java9ClassDefiner.java
+++ b/src/main/java/me/xdark/recaf/runtime/Java9ClassDefiner.java
@@ -32,7 +32,8 @@ final class Java9ClassDefiner implements ClassDefiner {
         try {
             MethodHandles.Lookup lookup = VMUtil.lookup();
             Class<?> unsafeClass = Class.forName("jdk.internal.misc.Unsafe", true, null);
-            MH_DEFINE_CLASS = lookup.findVirtual(unsafeClass, "defineAnonymousClass", MethodType.methodType(Class.class, Class.class, byte[].class, Object[].class));
+            MH_DEFINE_CLASS = lookup.findVirtual(unsafeClass, "defineAnonymousClass",
+                    MethodType.methodType(Class.class, Class.class, byte[].class, Object[].class));
             JDK_UNSAFE = lookup.findGetter(unsafeClass, "theUnsafe", unsafeClass).invoke();
         } catch (Throwable t) {
             throw new ExceptionInInitializerError(t);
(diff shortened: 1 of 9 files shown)


Col-E:
I'm kinda confused here. This seems to be more class-loading than simulation.  What does this do beyond the capabilities of [DefineUtil](https://github.com/Col-E/Recaf/blob/redesign/src/main/java/me/coley/recaf/util/DefineUtil.java)? 

The idea behind the #154 is for simulating potentially malicious code in a sandboxed environment. For instance, running a state-machine on a class file to record basic operations like primitives & strings on the stack/instance-fields but only not operating on potential vectors like file-writing/networking/etc.

xxDark:
The difference here is that defined classes will be GC'ed if they have no reference.
In fact, we still can make it to work as sandbox (For example, read `java.lang.String` as a resource, inject our code, and just define it (`defineClass` does not allow us to define `java.lang` classes outside of bootstrap class loader, but `defineAnonymousClass` does not perform any security checks)

src/main/java/me/xdark/recaf/runtime/util/VMUtil.java
```diff
@@ -0,0 +1,57 @@
...
+     * @return version of JVM
+     */
+    public static int vmVersion() {
+        return VM_VERSION;
+    }
+
+    static {
+        try {
+            VM_VERSION = (int) (Float.parseFloat(System.getProperty("java.class.version")) - 44);
+            Field field = Unsafe.class.getDeclaredField("theUnsafe");
```
andylizi: I've not tested it but wouldn't this fail on Java 9+ because there isn't a `sun.misc.Unsafe` anymore?

xxDark: Nope, but you are right, some methods were removed, see `Java9ClassDefiner`

src/main/java/me/xdark/recaf/runtime/Java9ClassDefiner.java
```diff
@@ -0,0 +1,42 @@
...
+        }
+    }
+
+    static {
+        try {
+            MethodHandles.Lookup lookup = VMUtil.lookup();
+            Class<?> unsafeClass = Class.forName("jdk.internal.misc.Unsafe", true, null);
+            MH_DEFINE_CLASS = lookup.findVirtual(unsafeClass, "defineAnonymousClass",
+                    MethodType.methodType(Class.class, Class.class, byte[].class, Object[].class));
+            JDK_UNSAFE = lookup.findGetter(unsafeClass, "theUnsafe", unsafeClass).invoke();
```
andylizi: Should be 
```suggestion
            JDK_UNSAFE = lookup.findStaticGetter(unsafeClass, "theUnsafe", unsafeClass).invoke();
```

andylizi: Oh and why not
```suggestion
            JDK_UNSAFE = lookup.findStatic(unsafeClass, "getUnsafe", MethodType.methodType(unsafeClass)).invoke();
```
by the way? There isn't any caller check, unlike `getUnsafe()` in the `sun.misc.Unsafe` version.

src/main/java/me/xdark/recaf/runtime/util/VMUtil.java
```diff
@@ -0,0 +1,57 @@
...
+            field.setAccessible(true);
+            Unsafe unsafe = UNSAFE = (Unsafe) field.get(null);
+            field = MethodHandles.Lookup.class.getDeclaredField("IMPL_LOOKUP");
+            MethodHandles.publicLookup();
```
andylizi: Unused line.

src/main/java/me/xdark/recaf/runtime/util/VMUtil.java
andylizi: Oh... I used to think that `sun.misc.Unsafe` was removed in Java 9, but apparently not...:sweat_smile:
They only moved it to the `jdk.unsupported` module and made it an wrapper for `jdk.internal.misc.Unsafe`.

andylizi: Oh... I used to think that `sun.misc.Unsafe` was removed in Java 9, but apparently not...:sweat_smile:
They only moved it to the `jdk.unsupported` module and made it an wrapper for `jdk.internal.misc.Unsafe`.

xxDark:
Well, I found out there is actually some restrictions in simulating runtime in this way.
Attach API must be used to redefine `java.lang` classes.
I can close this PR and re-implement stuff, or wait for more details in #154 
@Col-E

Col-E:
I updated the description of #154 - The intent should be more clear.

xxDark:
Alright, I'll close this PR and open new one soon (I hope so)

Issue #177, opened 2020-01-17
create a single folder for all output


blshkv:
Hi, 

recaf (1.15.9) writes all its output to the current directory instead of ~/.recaf or similar. That basically means that the entire system will be polluted with recaf folders and files which it is getting created during runtime.

Please consider to redesigning it.

Col-E:
Running Recaf shouldn't make too much stuff, 2 folders and 1 log file. :

![image](https://user-images.githubusercontent.com/21371686/72614112-6e813980-38ff-11ea-9bb5-f2c570976785.png)

I personally don't think that's bad. And the reason for it being the current directory instead of a specific folder is so that nobody has to ask _"Where does this config even save to?"_

xxDark:
Maybe settings should be stored in `~/.recaf` directory if user manually creates the directory first?

blshkv:
my point is that it is hard to trace back all these folders and files. "plugins" is just a common name, rc-config is may be openrc related (I have /usr/bin/rc-config binary in my system), txt file which is apparently a log file (must be .log) and even the binary (jar patched). All these files are getting created everytime in random folders. It would be hard to understand which software created after awhile and hard to clean a disk.

The last suggestion (having a directory) is tricky and it is like a hidden functionality. 
I would rather vote for a compilation flag if you really want it to make it optional

Col-E:
Fair enough. I've never had a setup using Recaf where the current directory wasn't where the executable was, so it never really occurred to me.

Col-E:
xxDark started on this, however the logging library used requires specifying a relative or absolute path in the config file _(Which wouldn't allow for platform independent paths)_

So if we wanted to move the log file to the home directory then we'd have to replace the log api. It's already mostly abstracted away to one class, so it could be replaced with a minimal one from scratch.

Col-E:
2.0.0 will be using `${user.home}/Recaf` from now on

blshkv:
Thank you guys!

PR #179, opened 2020-01-18
Cleanup code, prperly start JavaFX thread

xxDark:
## What's new
Start JavaFX thread properly
Recaf storage directory can be accessed with
`me.coley.recaf.Recaf.getStorageDirectory()`

## What's fixed
Solves #177

### DIFF
--- a/src/main/java/me/coley/recaf/config/Config.java
+++ b/src/main/java/me/coley/recaf/config/Config.java
@@ -1,12 +1,12 @@
 package me.coley.recaf.config;
 
 import com.eclipsesource.json.*;
+import org.apache.commons.io.FileUtils;
 
 import java.io.*;
 import java.lang.reflect.Field;
-import java.nio.file.Files;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
-import java.nio.file.StandardOpenOption;
 import java.util.*;
 
 import static me.coley.recaf.util.Log.*;
@@ -35,11 +35,8 @@ public abstract class Config {
 	}
 
 	@SuppressWarnings("unchecked")
-	void load(Path file) throws IOException {
-		final JsonObject json;
-		try (BufferedReader bufferedReader = Files.newBufferedReader(file)) {
-			json = Json.parse(bufferedReader).asObject();
-		}
+	void load(Path path) throws IOException {
+		JsonObject json = Json.parse(FileUtils.readFileToString(path.toFile(), StandardCharsets.UTF_8)).asObject();
 		for(FieldWrapper field : getConfigFields()) {
 			String name = field.key();
 			if(name == null)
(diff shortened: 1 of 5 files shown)


src/main/java/me/coley/recaf/config/Config.java
```diff
@@ -34,8 +35,11 @@ public String getName() {
 	}
 
 	@SuppressWarnings("unchecked")
-	void load(File file) throws IOException {
-		final JsonObject json = Json.parse(FileUtils.readFileToString(file, UTF_8)).asObject();
+	void load(Path file) throws IOException {
+		final JsonObject json;
+		try (BufferedReader bufferedReader = Files.newBufferedReader(file)) {
+			json = Json.parse(bufferedReader).asObject();
+		}
```
Col-E: Instead of this, the original can be used with only the small modification of `file` -> `file.toFile()`
Also, with the type change, the parameter name should change to something like `path`

src/main/java/me/coley/recaf/config/Config.java
```diff
@@ -120,8 +124,10 @@ else if(type.equals(List.class)) {
 				warn("Didn't write config for {}, unsure how to serialize.", name);
 		}
 		StringWriter w = new StringWriter();
-		json.writeTo(w, WriterConfig.PRETTY_PRINT);
-		FileUtils.write(file, w.toString(), UTF_8);
+		try (BufferedWriter bufferedWriter = Files.newBufferedWriter(file,
+				StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING)) {
+			json.writeTo(bufferedWriter, WriterConfig.PRETTY_PRINT);
+		}
```
Col-E: Same as above, can use `Path.getFile()` to keep the original one-liner

src/main/java/me/coley/recaf/control/gui/GuiController.java
```diff
@@ -19,7 +22,7 @@
  */
 public class GuiController extends Controller {
 	private final WindowManager windows = new WindowManager(this);
-	private final ConfigManager configs = new ConfigManager();
+	private final ConfigManager configs = new ConfigManager(Recaf.getStorageDirectory().resolve("rc-config"));
```
Col-E: Since this is the only usage of `getStorageDirectory()`, we may as well inline `resolve(rc-config)`. Plus it would make more sense in my opinion that the storage directory has the config folder defined locally.

src/main/java/me/coley/recaf/ui/MainWindow.java
```diff
@@ -145,18 +152,16 @@ public MainMenu getMenubar() {
...
-			window.controller = controller;
+            MainWindow app = window = new MainWindow(controller);
+            PlatformImpl.startup(() -> {
+                Stage stage = new Stage();
+                try {
+                    app.start(stage);
+                } catch (Exception ex) {
+                    throw new RuntimeException(ex);
+                }
+            });
```
Col-E: Using spaces, but the rest of the file is in tabs.

PR #183, opened 2020-01-19
Set log file to be generated at Recaf storage directory

xxDark:
## What's fixed
Solves #177 logging problem


Col-E:
Huh, I didn't see support for that in the tinylog documentation. Nice :+1:

PR #185, opened 2020-01-20
Code optimizations

xxDark:


src/main/java/me/coley/recaf/util/IOUtil.java
```diff
@@ -0,0 +1,104 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+/**
+ * Faster I/O utils
+ *
+ * @author xxDark
+ */
+public final class IOUtil {
```
Col-E: In general I think that refactoring out existing IO methods for this leads to overly verbose code. I would consider adding utility methods like these to simplify usage:
```java
// shorthand for common transfer calls
void transfer(InputStream in, OutputStream out)
    transfer(in, out, new byte[size], ANY)
// for cases like ShortcutUtil
void readBytes(InputStream in, int len)
   toByteArray(in, new ByteArrayOutputStream(), new byte[len], len)
```

src/main/java/me/coley/recaf/workspace/JarResource.java
```diff
@@ -29,6 +29,9 @@ public JarResource(File file) throws IOException {
 	@Override
 	protected Map<String, byte[]> loadClasses() throws IOException {
 		// iterate jar entries
+		ByteArrayOutputStream out = new ByteArrayOutputStream();
+		byte[] buffer = new byte[8192];
```
Col-E: For cases like this, how is the size of the buffer determined? I'm still curious as to why this is favorable over a one liner like `IOUtils.toByteArray(stream)`

src/main/java/me/coley/recaf/workspace/InstrumentationResource.java
```diff
@@ -33,17 +32,27 @@ public InstrumentationResource(Instrumentation instrumentation) {
...
+				in = ClassLoader.getSystemResourceAsStream(path);
+			} else {
+				in = loader.getResourceAsStream(path);
+			}
+			if (in == null) {
+				continue;
+			}
+			out.reset();
+			try (InputStream __ = in) {
+				classes.put(name, IOUtil.toByteArray(in, out, buffer));
```
Col-E: I think the control-flow can be optimized into this:
```java
ClassLoader loader = c.getClassLoader();
try(InputStream in = (loader != null) ? 
		loader.getResourceAsStream(path) : ClassLoader.getSystemResourceAsStream(path)) {
	if(in != null) {
		classes.put(name, IOUtils.toByteArray(in));
	}
}
```

src/main/java/me/coley/recaf/util/RegexUtil.java
```diff
@@ -101,11 +104,33 @@ public static Matcher getMatcher(String pattern, String text) {
...
+	 *
+	 * @param regex pattern's regex
+	 * @return {@link Pattern}
+	 */
+	public static Pattern pattern(String regex) {
+		return PATTERNS.computeIfAbsent(regex, Pattern::new);
+	}
+
+	/**
+	 * Checks if pattern matches an input.
```
Col-E: Should be `Checks if the entire input matches a pattern`

Since this is basically how `String.matches()` works, this checks if the entire length of the string matches. As opposed to the method above which will yield substring matches of the given text.

src/main/java/me/coley/recaf/workspace/JarResource.java
xxDark: Answered in Discord.

Issue #160, opened 2019-07-24
Russian characters in string are represent as \uXXXX


artemking4-Projects:
**Russian characters in string are represent as \uXXXX**

**To Reproduce**
1. Open any class that contains non-english characters
2. See \uXXXX instead of a character

Col-E:
Its much easier to modify unicode characters in this escaped representation. A lot of people can't easily create these characters on their machines. Plus then you don't have to worry that a person's machine doesn't support rendering the character _(font)_.

In a read-only context showing the unicode literal would make sense, for instance in the decompiled output. That's up to the decompiler how to print those characters though.

andylizi:
> Its much easier to modify unicode characters in this escaped representation

But how can you modify anything if you can't read it in the first place?

> A lot of people can't easily create these characters on their machines.

If you don't have the proper program to input those characters (e.g. an IME) then why would you try to *write* them. And you only need `Backspace` to *delete* them regardless of language anyway.
It's not like it's easy to write in `\u5929\u6c14\u4e0d\u9519`. Who, except a computer, can read or write in that?

> Plus then you don't have to worry that a person's machine doesn't support rendering the character _(font)_.

[A Unicode test page](http://www.madore.org/~david/misc/unitest/)
![](https://user-images.githubusercontent.com/12008103/61853683-1557c080-aeef-11e9-83c9-e6a119c1bb24.png)
![](https://user-images.githubusercontent.com/12008103/61853691-1983de00-aeef-11e9-9cf2-c03b60e1abdf.png)

Nope. Welcome to the 21st century.

Col-E:
I can accept when I'm wrong.

My mindset comes from working off of intentionally obfuscated code abusing unicode characters such as in the attached sample: [calc-obf.zip](https://github.com/Col-E/Recaf/files/3431167/calc-obf.zip)

Moving forward the option to use either presentation of international characters will be available, with the default for the literal values.

Col-E:
Using `StringEscapeUtils`  in the way shown in a81a41d6c5a38d24bab5590c3eb4aa2ee784687e will have to be changed. This was initially done to properly escape/unescape `\n` but this [also escapes international characters](https://stackoverflow.com/a/13700590), which is why we're here.

Col-E:
This has been resolved in 2.0 
 - #186
- 09075a904f59d8f0a5ac70d2e4daa690be0fd360

PR #186, opened 2020-01-20
Display unicode characters correctly

xxDark:
## What's fixed
Possiblle fix for #160

### DIFF
--- a/src/main/java/me/coley/recaf/ui/controls/text/TextPane.java
+++ b/src/main/java/me/coley/recaf/ui/controls/text/TextPane.java
@@ -1,5 +1,6 @@
 package me.coley.recaf.ui.controls.text;
 
+import com.github.javaparser.utils.StringEscapeUtils;
 import javafx.application.Platform;
 import javafx.geometry.Pos;
 import javafx.scene.Node;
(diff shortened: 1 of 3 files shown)


xxDark:
Ok so I see https://github.com/Col-E/Recaf/issues/160#issuecomment-515003949, so I'll think how we can workaround this.

Col-E:
Yeah, its a tricky matter of balancing what _should be_ escaped and what _should not be_ escaped. Looking forward to seeing what you come up with.

xxDark:
Probably solves #160

xxDark:
poc

Col-E:
Can you drop a test case or few into the `AssemblyAstTest` test class with some examples? I'd like to have a test or two to flesh out how we want unicode support to work. It'll also allow us to verify we don't revert support accidentally if we change things in the future.

Col-E:
I asked for cases to be made in `AssemblyAstTest` not to verify that this works, but to verify that it integrates with the AST parsers. So for instance:
```java
// chinese
LDC "下雨了"
// chinese escaped
LDC "\u4E0B\u96E8\u4E86"
```
Then we would assert that the `LdcInsnAst.value` is `下雨了` _(and not the literal value of the escaped unicode patterns)_

Col-E:
I'll merge this and ensure this works for LDC strings. But for type/descriptor parsing, we'll handle that hurdle when we get there.

PR #189, opened 2020-01-21
WIP: Simulation

xxDark:
## What's new

Inital concept for #154 
As of now this branch wont even compile.


Col-E:
Looking forward to seeing what this will turn into. I think handling object types will be a real challenge, like for example:
```java
String s = "Hello World";
s = s.substring(6);
```
And we'd know `s` is now `"World"`.

I was thinking for core classes that are safe to use in all regards _(Like String)_ we can just wrap literal instances of them. But for things like networking, where we wouldn't actually want to download anything, we would use a sort of mocking system to simulate dummy instances. 

There's a lot of ways we can handle it, what do you think makes sense?

PR #200, opened 2020-04-24
Fix Chinese translation

Enaium:
Fix Chinese translation


andylizi:
These should be merged to `redesign` and not `master`, as many keys don't exist in the `en.json` of `master`.

应该合并到 redesign 分支去，创建PR时选错了。

Issue #211, opened 2020-05-07
Show filename in the window caption


pedroflor:
## Show filename in the window caption

Could be possible to show in the window the name of the file opened?
![image](https://user-images.githubusercontent.com/3332674/81349001-22fcc200-908d-11ea-9a82-8f8949f27b6a.png)

xxDark:
@pedroflor ?
![poc](https://user-images.githubusercontent.com/19853368/82652376-2f088800-9c26-11ea-9f49-718a5ad46bc6.png)

Col-E:
Latest

pedroflor:
Great!
Thanks again.

pedroflor:
> @pedroflor ?
> ![poc](https://user-images.githubusercontent.com/19853368/82652376-2f088800-9c26-11ea-9f49-718a5ad46bc6.png)

Thanks for your time!

PR #216, opened 2020-05-22
Features/fixes/additions.

xxDark:
## What's new
1. Add #211 
2. Added support for resource locations.
3. Automatically load `attach` library when possible.
4. Allow plugins to intercept workspaces

## What's fixed

1. Recaf idles in background
2. `File` is replaced with `Path`, allowing to use more file systems if needed
3. Replacing `File` with `Path` is backwards compatible.
4. Javac used ';' separator for classpath, which is not the correct case on Linux systems.


src/main/java/me/coley/recaf/workspace/JavaResource.java
```diff
@@ -352,26 +354,26 @@ public void invalidate() {
...
 	 *
 	 * @return Map of class names to their documentation.
 	 *
 	 * @throws IOException
 	 * 		When the file could not be fetched or parsed.
 	 */
-	protected Map<String, Javadocs> loadDocs(File file) throws IOException {
-		Map<String, Javadocs> map = new HashMap<>();
+	protected Map<String, Javadocs> loadDocs(Path path) throws IOException {
+		Map<String, Javadocs> map = new HashMap<>(16, 1F);
```
Col-E, line 374: I don't see why specifying the initial capacity or load factor is necessary. 
Most javadoc archives have much more than 16 files too.

src/main/java/me/coley/recaf/workspace/WorkspaceIO.java
```diff
@@ -17,6 +20,19 @@
  * @author Matt
  */
 public class WorkspaceIO {
+	/**
+	 * @param json
+	 * 		Json path.
```
Col-E, line 25: See `path` should be `file` in comment as mentioned before.

Col-E:
Still some missing javadoc fixes but I'll get those in a follow up commit. Tested locally, I wanted to test paths not changing behavior in headless mode but found that it crashes instead:

```
Caused by: java.lang.NoSuchMethodException: me.coley.recaf.command.completion.FileCompletions.<init>()
        at java.lang.Class.getConstructor0(Class.java:3082)
        at java.lang.Class.getDeclaredConstructor(Class.java:2178)
        at picocli.CommandLine$DefaultFactory.create(CommandLine.java:4888)
        at picocli.CommandLine$DefaultFactory.create(CommandLine.java:4911)
```

The `FileCompletions` class requires a default constructor. Something else I'll fix up. Perhaps the cli should be tested on some level to make cases like this more clear.

Col-E:
Also, I appreciate the `@Deprecated` copies for the old file way, but since 2.0.0 is still in pre-release I don't consider it necessary. I've gone and removed them, replacing any usage with the NIO alternative.

Col-E:
Oh also in `Natives`, on my system when loadLibrary is called when attaching is already correctly configured it breaks attaching.

The attach provider is no longer configured. So I added a check for that. Perhaps this can be looked into futher and tested to see if there are any other odd cases like this.

PR #225, opened 2020-06-19
fix: remove outdated JavaParser code

xxDark:
## What's fixed

This issue was fixed in 3.15.17


Col-E:
It's been a long time since I made that class because of whatever the problem was, but I recall it was due to something in RichTextFX, not JavaParser. I have no clue what the issue was, what triggered it, and if it's actually been fixed.

Col-E:
Commenting to ping regarding prior commit. 
I don't see anything in their change-logs regarding the problem.

xxDark:
> 
> 
> Commenting to ping regarding prior commit.
> I don't see anything in their change-logs regarding the problem.

https://github.com/javaparser/javaparser/issues/2437

Col-E:
Hmmm... If thats what propagated up and caused the UI to spaz out that's pretty interesting. I'll see if I can reproduce the old behavior and swap out with these changes. Will merge after a short bit of testing.

PR #229, opened 2020-07-02
Keep scroll position after reloaded.

### LINKED ISSUE #228: Scroll position not kept when renaming
**Describe the bug**

When a field, method, or class is renamed, the decompiler view is reset and the scroll position goes back to the start.

**To Reproduce**

1. Open any jar file
2. Navigate to any class
3. Rename a field or method
4. The decompiler pane will reload and the scroll position will be back on top

TimmyOVO:
## What's fixed

* Fixed  #228


Col-E:
Looks good to me, and build succeeds 👍 

I'll merge this soon

PR #231, opened 2020-07-06
fix: NPE if the workspace not initialized

sq179:
## What's new

* Additional check for null in ```me.coley.recaf.ui.MainMenu#addLibrary(...)```

## What's fixed

* ```NullPointerException``` when add library with empty workspace (doesn't loaded jar file):

image

### DIFF
--- a/src/main/java/me/coley/recaf/ui/MainMenu.java
+++ b/src/main/java/me/coley/recaf/ui/MainMenu.java
@@ -1,7 +1,9 @@
 package me.coley.recaf.ui;
 
 import javafx.scene.Node;
-import javafx.scene.control.*;
+import javafx.scene.control.Menu;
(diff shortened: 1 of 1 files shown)


src/main/java/me/coley/recaf/ui/MainMenu.java
```diff
@@ -235,10 +237,16 @@ private void addLibrary() {
 		fcLoadApp.setInitialDirectory(config().getRecentLoadDir());
 		List<File> files = fcLoadApp.showOpenMultipleDialog(null);
```
xxDark, line 238: Why don't you add check before opening file dialog?

sq179: @xxDark, oh, i didn't think about it at all, i'll change location of check with new commit

Issue #220, opened 2020-06-10
Config being incorrectly written, causing first-time prompt to always show


Col-E:
For some reason config load/save is called twice, and the second write overwrites the config with the wrong values.
```
10:01:28.088 [main] INFO : Recaf-2.0.1
10:01:28.789 [main] INFO : Discovered plugin 'String Replace-1.0.0'
10:01:28.789 [main] INFO : Discovered plugin 'Theme Editor-1.0.0'
10:01:28.790 [main] INFO : Loaded 2 plugins
10:01:29.383 [main] TRACE: Loading configuration
10:01:31.922 [main] TRACE: Loading configuration
10:01:35.767 [JavaFX Application Thread] TRACE: Exiting FX platform
10:01:35.808 [JavaFX Application Thread] INFO : Shutting down
10:01:35.823 [Thread-8] TRACE: Saving configuration
10:01:35.823 [Thread-4] TRACE: Saving configuration
```

PR #237, opened 2020-08-03
fix: controller initializes before plugin manager loads plugins

xxDark:


Col-E:
Didn't we originally have the load call here and move it down for some reason?

https://github.com/Col-E/Recaf/commit/bb864ad57978f1f471089d47b8dad155da795c36#diff-304402589558d82a00b223b7edf664d5

Col-E:
Going to close this later since this will revert the fix for #220

Col-E:
Perhaps instead we can add some additional logic to do the custom-command additions someplace later in the execution instead of where it is currently handled.

Issue #245, opened 2020-08-23
Compile snippets into bytecode in assembler


Col-E:
## Compile snippets into bytecode in assembler

The idea is to allow users to open a method assembler, then open a prompt allowing them to type in something like `System.out.println("test")`. This line would then be compiled against the current source and inserted into the assembler as lines of method bytecode.

It would be nice to support arbitrary variable support too, such as: `if ($0 != 0){} ` --> `*LOAD 0`

Crystallinqq:
I would shit my pants if this happened.

andylizi:
> I would shit my pants if this happened.

[Javassist](https://www.javassist.org/tutorial/tutorial2.html) kinda already has this feature, up to and including variables support:
```java
class Point {
    int x, y;
    void move(int dx, int dy) { x += dx; y += dy; }
}
```
```java
ClassPool cp = ClassPool.getDefault();
CtClass cc = cp .get("Point");
CtMethod m = cc.getDeclaredMethod("move");
m.insertBefore("{ System.out.println(\"dx = \" + $1 + \", dy = \" + $2); }");
Class c = cc.toClass();
Point p = (Point) c.newInstance();
p.move(4, 2);
```

xxDark:
> 
> 
> > I would shit my pants if this happened.
> 
> [Javassist](https://www.javassist.org/tutorial/tutorial2.html) kinda already has this feature, up to and including variables support:
> 
> ```java
> class Point {
>     int x, y;
>     void move(int dx, int dy) { x += dx; y += dy; }
> }
> ```
> 
> ```java
> ClassPool cp = ClassPool.getDefault();
> CtClass cc = cp .get("Point");
> CtMethod m = cc.getDeclaredMethod("move");
> m.insertBefore("{ System.out.println(\"dx = \" + $1 + \", dy = \" + $2); }");
> Class c = cc.toClass();
> Point p = (Point) c.newInstance();
> p.move(4, 2);
> ```

We discussed that on Discord, however, there are some problems:
How can we use values from the stack without registers?
Javassist does not support ASM, so all the output must be converted.

andylizi:
> How can javassist access the value on the stack?

No it can't - at least, not in its current state. After some digging however, I believe its possible to make it work without major modifications.

[...]

PR #249, opened 2020-08-25
feat: Extend API for AST.

xxDark:
## What's new

Extended API for AST parsing. Required for #245


Col-E:
Thanks for fixing my shitty duck-taped API with a proper design ❤️  

Reviewing and merging in a bit

Issue #274, opened 2020-09-12
Illegal named methods or variable names


MizzMaster:
You know that there are words that we cant use while declaring a method or a variable. BUT FU***NG OBFUSCATORS CAN!
I spent more than 30 hours dealing with bugs. I got mad, sorry.

If an obfuscator rename a variable with an illegal(unicode or reserved keywords), this causes Recaf suck while parsing decompiled code. If an variable name gets invisible, that means it is unicode, but if a method name gets unparseable, it means you cant even change it's name using table mode.

In my jar, there are 2 methods named "do" and "if" and maybe a lot more. I don't know if it is your job to fix this but it causes your program to cant even rename this illegal method. If you say that is not my job, i will open an issue on "Java Deobfuscator" repo.

Btw, you can run CFR with -renameillegalidents to prevent unicode bug.

andylizi:
`1.x` versions of Recaf allowed changing all CFR options. It'd be nice if this can be supported again. Some reflection (e.g. [on here](https://github.com/leibnitz27/cfr/blob/00f0a883e8e3423a0e3d24930a37560bfbfbdf34/src/org/benf/cfr/reader/util/getopt/OptionsImpl.java#L266-L539)) can be used to avoid the tedious work of keeping all the options up-to-date.

`--renameillegalidents` would solve the problem in this case.

MizzMaster:
Does --renameillegalidents change the ones that use reserved keywords?

andylizi:
`$ cfr --help renameillegalidents`
> Rename identifiers which are **not valid java identifiers**.  Note - this WILL break reflection based access, so is not automatically enabled.

Yes.

MizzMaster:
I hope this does not break Recaf

MizzMaster:
CFR just decompiles and gives us text based .class file. Not all the time perfect and not all the time same, i was trying to rename that illegal metod in Recaf but i couldnt

So, CFR doesnt change the actual name, it just changes while showing you up

MizzMaster:
And i forgot to mention, I was trying to rename it on "Table Mode", but when i press enter, it does nothing

andylizi:
~If you just want to rename stuff, the Table view will surely work?~

Or not, I guess.

---

By the way, I've always find the decompile view not very helpful because any obfuscation can easily break the parser, and if that failed, the whole page is reduced to a fancy textbox. Proper IDEs and language servers have the ability to parse partial source, but I imagine that'd be hard to implement.

[...]

PR #287, opened 2020-09-18
feat: Support parsing error recovery

andylizi:
After [this conversation](https://github.com/Col-E/Recaf/issues/274#issuecomment-691491953) I've been doing some research. It seems that JavaParser [already supports some level of error recovery](https://github.com/javaparser/javaparser/pull/952), but currently there're two problems with this:
1. Recaf throws out the parse result immediately when it contains any error.
2. The kinds of recovery JavaParser can do are limited, especially when the class structure is compromised. This unfortunately happens a lot with obfuscated code in CFR.

For the first problem, we can instead check `CompilationUnit.getParsed()` against `Parsedness.PARSED`. The result in such case may not be fully formed, but after the patch on `JavaParserUtil.checkForDeclaredSelection()` it seems to be working without problem.

For the second one, we can try to comment out lines that could be causing us problem and use the result under-the-hood. There aren't any symbols to be parsed in comments, so the resulting AST should still correctly map to the original source. Furthermore, this workaround will only be used when parsing haven't succeed at least once and there ain't any info for context actions yet. This is 
 in part to prevent it from giving out inconsistent result while users are modifying the code. It isn't the most graceful of solutions, but it seems to work well enough in my testing.

### DIFF
--- a/src/main/java/me/coley/recaf/parse/source/SourceCode.java
+++ b/src/main/java/me/coley/recaf/parse/source/SourceCode.java
@@ -6,6 +6,7 @@ import com.github.javaparser.ast.Node;
 import com.github.javaparser.ast.body.TypeDeclaration;
 import com.github.javaparser.ast.expr.*;
 import com.github.javaparser.ast.nodeTypes.NodeWithName;
+import me.coley.recaf.util.JavaParserUtil;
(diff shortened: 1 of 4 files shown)


Col-E:
Merged locally with some minor rebasing, github will show as closed though.

Issue #281, opened 2020-09-17
Record inner mappings & allow exporting applied mappings


Col-E:
## Mapping tracking

Recaf does not currently record the history of mapping actions applied to classes. Tracking these changes and allowing users to export the sequence of changes as a mappings file would be useful. Especially if the exported mappings could be transformed into any arbitrary format.

Ideally, the read-only mapping utilities should be rewritten to support writing as well. The usage of these classes via `MappingsImpl` will need to be refactored to support being more than reader utilities.

Col-E:
#295 gives an excellent start to this, the remaining part is to map the recorded ASM format mappings to other formats.

As mentioned, this will require some refactoring...

Col-E:
3X will allow exporting to any of the supported mappings
![java_mpaPOgnaDT](https://user-images.githubusercontent.com/21371686/163449730-3bef2a21-76dd-4d73-b4ca-c79d8b47c346.gif)

PR #295, opened 2020-10-03
feat: Track applied renames in workspace and enable mapping export

RagingCactus:
First, thanks for creating this project! This could have helped me a lot with a project some years ago, so I took a look at partially implementing #281.

## What's new
The workspace now keeps track of mappings applied to it in an aggregated map conforming to the internal ASM format. This aggregated map can then be used to provide a basic export facility for mappings in that format (the Simple format).

The mapping in the workspace only stores the end result of all applied mappings. That means that rename operations "a -> b -> c" are folded together and stored as "a -> c". To enable this, `AsmMappingUtils` contains code to apply new mapping rules to existing mappings by just operating on the mapping Strings itself rather than a real class graph.

While it would be nice to be able to export to all the supported import formats, this is a start and should be possible to expand on later.

This is a partial implementation of #281

## Implementation Notes
I haven't worked too much with ASM and remapping before, so I hope my very very primitive approach to handling transitive mappings is not fundamentally flawed. Please educate me if it is though! 😄 

There are some especially rough edges when handling constructors, initializers and invokedynamic remapping in `AsmMappingUtils`. Can you point me to a resource or test case to understand these better?

### DIFF
--- a/src/main/java/me/coley/recaf/config/ConfBackend.java
+++ b/src/main/java/me/coley/recaf/config/ConfBackend.java
@@ -39,6 +39,11 @@ public class ConfBackend extends Config {
 	@Conf("backend.recentsave")
 	public String recentSaveWorkspace = CURRENT_DIR;
 	/**
+	 * Recent path used by the save-map dialog.
+	 */
+	@Conf("backend.recentsave")
+	public String recentSaveMap = CURRENT_DIR;
+	/**
 	 * Check to determine if user should be told to read the documentation.
 	 */
 	@Conf("backend.firsttime")
@@ -103,6 +108,14 @@ public class ConfBackend extends Config {
 	 */
 	public File getRecentSaveWorkspaceDir() {
 		return getDir(recentSaveWorkspace);
+	}
+
+	/**
+	 * @return Directory to use for save-map file-chooser.
+	 * Based on the most recent file exported.
+	 */
+	public File getRecentSaveMapDir() {
+		return getDir(recentSaveMap);
 	}
 
 	private File getDir(String root) {
--- a/src/main/java/me/coley/recaf/config/ConfBackend.java
+++ b/src/main/java/me/coley/recaf/config/ConfBackend.java
@@ -31,17 +31,17 @@ public class ConfBackend extends Config {
 	/**
 	 * Recent path used by the save-application dialog.
 	 */
-	@Conf("backend.recentsave")
+	@Conf("backend.recentsave.app")
 	public String recentSaveApp = CURRENT_DIR;
 	/**
 	 * Recent path used by the save-workspace dialog.
 	 */
-	@Conf("backend.recentsave")
+	@Conf("backend.recentsave.workspace")
 	public String recentSaveWorkspace = CURRENT_DIR;
 	/**
 	 * Recent path used by the save-map dialog.
 	 */
-	@Conf("backend.recentsave")
+	@Conf("backend.recentsave.map")
 	public String recentSaveMap = CURRENT_DIR;
 	/**
 	 * Check to determine if user should be told to read the documentation.
--- a/src/main/java/me/coley/recaf/mapping/AsmMappingUtils.java
+++ b/src/main/java/me/coley/recaf/mapping/AsmMappingUtils.java
@@ -1,0 +1,161 @@
+package me.coley.recaf.mapping;
+
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimap;
+import com.google.common.collect.Multimaps;
+
+import java.util.AbstractMap;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+/**
+ * Util class to work with ASM mappings based on String operations, such as extracting portions of ASM mapping keys or
+ * folding new mappings onto an existing mapping.
+ */
+public class AsmMappingUtils {
+    private AsmMappingUtils() {
+        // static class
+    }
+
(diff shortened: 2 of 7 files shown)


src/main/java/me/coley/recaf/config/ConfBackend.java
```diff
@@ -38,6 +38,11 @@
 	 */
 	@Conf("backend.recentsave")
 	public String recentSaveWorkspace = CURRENT_DIR;
+	/**
+	 * Recent path used by the save-map dialog.
+	 */
+	@Conf("backend.recentsave")
```
Col-E, line 44: The config identifier should be different here so it doesn't conflict with the workspace one above

src/main/java/me/coley/recaf/mapping/AsmMappingUtils.java
```diff
@@ -0,0 +1,161 @@
...
+            return null;
+
+        boolean isMember = key.contains(".");
+        if (!isMember) {
+            // This is a class, just return the original mapping as its value is the applied value
+            return mapping;
+        }
+
+        // Don't map invokedynamic calls
+        boolean isInvokeDynamic = key.charAt(0) == '.';
```
Col-E, line 144: Just curious, what cases does this come up for?

src/main/java/me/coley/recaf/mapping/AsmMappingUtils.java
```diff
@@ -0,0 +1,161 @@
...
+import java.util.AbstractMap;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+/**
+ * Util class to work with ASM mappings based on String operations, such as extracting portions of ASM mapping keys or
+ * folding new mappings onto an existing mapping.
```
Col-E, line 16: Since you wrote this class feel free to add yourself to the docs with `@author`

Col-E (changes requested):
Great stuff all around 👍 
Just one minor change and I'll merge

src/main/java/me/coley/recaf/mapping/AsmMappingUtils.java
RagingCactus: This is where my (somewhat limited) understanding of the remapping (and details of invokedynamic) ends. As stated in a comment earlier I was looking at `SimpleRecordingRemapper.map()` in order to not miss any case and that function is prepared to handle these. I did not want to just pretend the case can't exist.
(This is where the strict response to your comment stops and speculation starts)

Unfortunately I have no idea how this would be handled correctly. It certainly looks like it doesn't need to just for reading code. I suspect that the remapped classes in memory are very messed up right now. I did a small test with an interface and a lambda:

```java
@FunctionalInterface
public interface StringSupplier {
    String getString();
}
```
```java
public class Main {
    public static void main(String[] args) {
        StringSupplier supp = () -> "Hello!";
        System.out.println(supp.getString());
    }
}
```

I threw a jar with these two files into Recaf and renamed `StringSupplier.getString()` to `StringSupplier.getRenamedString()`. I then opened the disassembly for `main()`. This showed the following bytecode disassembly:
```
DEFINE PUBLIC STATIC main([Ljava/lang/String; args)V
ALIAS H_META "handle[H_INVOKESTATIC java/lang/invoke/LambdaMetafactory.metafactory(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;Ljava/lang/invoke/MethodType;Ljava/lang/invoke/MethodHandle;Ljava/lang/invoke/MethodType;)Ljava/lang/invoke/CallSite;]"
A:
LINE A 3
INVOKEDYNAMIC getString ()LStringSupplier; ${H_META} args[()Ljava/lang/String;, handle[H_INVOKESTATIC Main.lambda$main$0()Ljava/lang/String;], ()Ljava/lang/String;]
ASTORE supp
B:
LINE B 4
GETSTATIC java/lang/System.out Ljava/io/PrintStream;
ALOAD supp
INVOKEINTERFACE StringSupplier.getRenamedStr()Ljava/lang/String;
INVOKEVIRTUAL java/io/PrintStream.println(Ljava/lang/String;)V
C:
LINE C 5
RETURN
D:
```

To me it looks like the `LambdaMetafactory` will receive the old method name argument from `INVOKEDYNAMIC getString ()LStringSupplier;` while the invokeinterface instruction has been changed to the new name.
Again, this doesn't seem to bother the decompilers, but I guess this would be the case where remapping invokedynamic comes up.

I'll stress again that my understanding of the JVM and bytecode instructions is rather limited. This test was just a product of my curiosity.

src/main/java/me/coley/recaf/config/ConfBackend.java
RagingCactus: I thought this was intentional as `recentSaveApp` and `recentSaveWorkspace` already use the same identifier.
I'll change all 3, guessing it was an accident all along. Feel free to request more changes if that isn't what you want.

src/main/java/me/coley/recaf/mapping/AsmMappingUtils.java
RagingCactus: I personally don't think `@author` provides much value compared to what `git blame` and `git log` can do, so unless you want me to I'll pass

src/main/java/me/coley/recaf/config/ConfBackend.java
Col-E: Oh crap. Looks like I've missed something all along.

Col-E (approved):

PR #304, opened 2020-11-04
macOS experience improvements

Anfanik:
## What's new

* Allow config to use system menubar
* Remove window icon
* Add dock icon

### DIFF
--- a/src/main/java/me/coley/recaf/config/ConfDisplay.java
+++ b/src/main/java/me/coley/recaf/config/ConfDisplay.java
@@ -57,6 +57,11 @@ public class ConfDisplay extends Config {
 	 */
 	@Conf("display.maxrecent")
 	public long maxRecentFiles = 6;
+	/**
+	 * Use system menubar if it is possible.
+	 */
+	@Conf("display.usesystemmenubar")
+	public boolean useSystemMenubar;
 
 	ConfDisplay() {
 		super("display");
--- a/src/main/java/me/coley/recaf/config/ConfDisplay.java
+++ b/src/main/java/me/coley/recaf/config/ConfDisplay.java
@@ -58,7 +58,7 @@ public class ConfDisplay extends Config {
 	@Conf("display.maxrecent")
 	public long maxRecentFiles = 6;
 	/**
-	 * Use system menubar if it is possible.
+	 * Use system menubar on macOS.
 	 */
 	@Conf("display.usesystemmenubar")
 	public boolean useSystemMenubar;
--- a/src/main/java/me/coley/recaf/ui/MainMenu.java
+++ b/src/main/java/me/coley/recaf/ui/MainMenu.java
@@ -21,6 +21,7 @@ import me.coley.recaf.ui.controls.*;
 import me.coley.recaf.util.ClasspathUtil;
 import me.coley.recaf.util.IOUtil;
 import me.coley.recaf.util.Log;
+import me.coley.recaf.util.OSUtil;
 import me.coley.recaf.util.self.SelfUpdater;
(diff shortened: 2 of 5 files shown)


Col-E:
I'm not too familiar with Macs, what does this look like when used?

src/main/java/me/coley/recaf/ui/controls/ConfigPane.java
```diff
@@ -48,6 +48,7 @@ public ConfigPane(GuiController controller, ConfDisplay config) {
 		editorOverrides.put("display.forceWordWrap", Toggle::new);
 		editorOverrides.put("display.suggest.classerrors", Toggle::new);
 		editorOverrides.put("display.maxrecent", v -> new NumberSlider<>(controller, v, 0, 20, 2));
+		editorOverrides.put("usesystemmenubar", Toggle::new);
```
Col-E, line 51: Should be `display.usesystemmenubar`.

src/main/resources/translations/en.json
```diff
@@ -497,6 +497,8 @@
 	"display.suggest.classerrors.desc": "Show suggestion popups when certain errors occur",
 	"display.maxrecent.name": "Max recent files",
 	"display.maxrecent.desc": "How many items to show in recent files menu.",
+	"display.usesystemmenubar.name": "System menubar",
+	"display.usesystemmenubar.desc": "Use system menubar if it is possible.",
```
Col-E, line 501: Can add some additional context, perhaps something like `Use system menubar on MacOS`

src/main/java/me/coley/recaf/ui/MainMenu.java
```diff
@@ -70,7 +71,12 @@
 	public MainMenu(GuiController controller) {
 		// TODO: Properly managed disabled state of menu items
 		this.controller = controller;
-		//
+
+		boolean isUseSystemMenuBar = controller.config().display().useSystemMenubar;
+		// macOS menubar don't show empty items
+		boolean isEmptyMenuItemsSupported = !isUseSystemMenuBar || OSUtil.getOSType() != OSUtil.MAC;
+		setUseSystemMenuBar(isUseSystemMenuBar);
```
Col-E, line 78: I think It would be more clear what the intent is if the logic was: _"Not using the menu bar, but if we are using the bar then we must be on mac"_

> `!bar || (bar && mac)`

You can ignore this, just talking out loud here

src/main/java/me/coley/recaf/config/ConfDisplay.java
```diff
@@ -57,6 +57,11 @@
 	 */
 	@Conf("display.maxrecent")
 	public long maxRecentFiles = 6;
+	/**
+	 * Use system menubar if it is possible.
```
Col-E, line 61: As mentioned below, may want to mention that this is a mac-only feature in the comment here.

Col-E (changes requested):
Just a minor typo in the config pane and possibly clarifying that this is a mac-only feature in the comments. Looks good otherwise 👍

Anfanik:
> I'm not too familiar with Macs and I don't have one to test on. What does this look like when used?

image

src/main/resources/translations/en.json
Anfanik: I made a description like this because as I know, some Linux environments have a system menubar too, but I'm not sure.  😖

Col-E (approved):

Col-E:
Merged into `2.13.0` 👍 

I made it so the mac-only option does not show up on windows/linux user's config window

PR #308, opened 2020-12-07
feat: Resource Interceptor Plugin

1fxe:
## What's new

Plugin intercepts resource loading, allowing you to load a jar (or other available resources) without having to rename the original file to the correct extension.
[Example](https://github.com/filipwtf/Recaf-Lunar/blob/master/src/main/java/wtf/filip/lunar/LunarPlugin.java)

## What's fixed
N/A

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,11 +9,11 @@
 	<!-- Variables -->
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		<asm.version>9.0</asm.version>
+		<asm.version>9.1</asm.version>
 		<analysis.version>1.6.0</analysis.version>
 		<dude.version>1.1.1</dude.version>
 		<cfr.version>0.151</cfr.version>
-		<ff.version>401</ff.version>
+		<ff.version>403</ff.version>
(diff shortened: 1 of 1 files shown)


Col-E:
A closer name would be a ResourceExtensionInterceptor.

If we wanted to give full access to what a user is loading to plugins we'd want to also provide and allow modification of:

- Assumed extension _(What this PR does so far)_
- Assumed name _(Could allow it to be shown as something different in the UI)_
- Pointer to content
    - Kind of content _(jar, war, class, directory, agent, etc)_
    - File/Path if relevant _(Recaf already supports shortcuts as inputs, but providing this info to plugins would be nice)_

filipwtf:
> * Assumed name _(Could allow it to be shown as something different in the UI)_

I'm not really sure how to edit the UI however I added the  ability to return an assumed name.

zanins:
Go

zanins (approved):

king5300:
Javax

king5300:
Babs

king5300:
[]()

king5300:
> Javax

king5300:
Java

Issue #309, opened 2020-12-09
Handle multiple members with same name when renaming


mmiszczyk:
## Handle multiple members with same name

Some obfuscators (e.g. Proguard) can rename multiple variables of different types to a same name. You can see examples of this if you load [the DoomRPG mobile game available on archive.org](https://archive.org/details/DoomRPG) into Recaf. For example: if you load it and open the ```DoomRPG``` class, you'll notice that all the member variables are named ```a```.
```java
    public static DoomRPG a;
    public static k a;
    public static Random a;
    public static Display a;
    public static Runtime a;
    public static long a;
    public boolean a;
```
This makes the decompiled code uncompileable (JVM bytecode allows multiple variables with one name, Java language standard does not), which isn't ideal but it's to be expected: decompilers are heuristics-based, and there is no guarantee that they'll give you something that works.

Bigger problem is that this really confuses Recaf. If I open the class in table view, go into the Fields tab and start renaming, it will rename all the fields. This means that I can't give everything a readable name like calling ```Random a``` ```Random rand```, as it will turn everything into ```rand``` (```Display rand```, ```Runtime rand``` etc.).

Currently, the only way to do the renaming is by editing assembly. Unfortunately, this is a purely manual process. I can do this:
```jasmin
DEFINE PUBLIC STATIC Ljava/util/Random; rand
```
but I'll still have code like:
```jasmin
INVOKESPECIAL java/util/Random.<init>()V
PUTSTATIC DoomRPG.a Ljava/util/Random;
```
or, in decompiled Java:
```java
            a = new Random();
```
so I need to find all the usages of ```a``` as an instance of ```Random``` and manually rename them to ```rand```. This is very tedious and error-prone.

CFR decompiler is able to handle this through a ```--renamedupmembers``` command-line flag, so being able to enable it would mitigate the problem a little bit. It wouldn't be a real solution though, as CFR isn't always able to decompile code so that it works in exactly the same way as the original, or even works at all (to continue with the example used here: CFR is unable to give any sort of output for the aforementioned ```DoomRPG``` class, and some other classes from the same file - e.g. the ```k``` class - will contain pseudocode with gotos, which will not be compiled by ```javac```).

The real solution would be for Recaf to not assume that one name always refers to the same field. When inspecting the class with Table view, it clearly recognizes that those are 7 different variables - it only treats them as one when renaming.

Col-E:
[...]

PR #310, opened 2020-12-09
Handle none/ package

mmiszczyk:
## What's fixed

When Enigma sees a class without a packagename (which seems to be common for JavaME MIDlets, e.g. DoomRPG mentioned in #309), it puts it inside a dummy package 'none/'. Recaf does not do that, so Enigma mappings for those package-less classes don't work when imported into Recaf. I've been dealing with it by removing all instances of string ```none``` with ```sed```, but it would be nice to do this inside Recaf's parser.

### DIFF
--- a/src/main/java/me/coley/recaf/mapping/EnigmaMappings.java
+++ b/src/main/java/me/coley/recaf/mapping/EnigmaMappings.java
@@ -35,7 +35,7 @@ public class EnigmaMappings extends FileMappings {
 	}
 	
 	private static String removeNonePackage(String text){
-		return text.replaceAll("(^|L)none/", "");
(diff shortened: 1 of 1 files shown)


Col-E:
Fun fact, you can slightly increase performance by using non-capturing regex groups. If the mappings are pretty big lots of small improvements add up. 

Overall, looks good and I'll get to this later today most likely

mmiszczyk:
I'll be honest with you, I am unable to write a good regex without using a visualizer. So I'll leave performance improvements to more competent people.

Col-E:
Its a small 2 character change, but I'll add it on after I merge this locally. It'll show up as merged when I push later.

PR #320, opened 2020-12-28
Pull UI code into UI module

Col-E:
Ideally the core module will not have any of the `Presentation` implementations for the UI. The default `EmptyPresentation` will be included for the core, and that should be it.

So, if the core module defaults to using the empty presentation, the ui module will have to override that when invoking the core. Gonna need to do command line parsing for that if we insert the presentation type as an argument.


Col-E:
> Can't we just use method references here?

Not if they're referenced in the core, but defined in the UI module. 

We don't want the core to depend on the UI module.

xxDark:
> 
> 
> > Can't we just use method references here?
> 
> Not if they're referenced in the core, but defined in the UI module.
> 
> We don't want the core to depend on the UI module.
> 
> Reflection in this case is to allow this to separate the work without compile errors.

I think that we will shoot ourselves in the leg some time later if we do reflection hacks, can't we workaround that somehow?

Col-E:
Its still a WIP branch, and I'm all ears. 

I'd like to do the `::new` but retain module independence for `core` from `ui`

recaf-ui/build.gradle
```diff
@@ -0,0 +1,11 @@
+plugins {
+    id 'application'
+}
+
+dependencies {
+    implementation project(':recaf-core')
+}
+
+application {
+    mainClass = 'me.coley.recaf.RecafUI'
```
xxDark, line 10: We can also keep that simple.

```Groovy
jar {
    manifest {
        attributes('Main-Class': 'me.coley.recaf.RecafUI')
    }
}
```
No need for 3d-party plugins.

Col-E: I'm no gradle expert. But then if you try and run it in IntelliJ it doesn't actually run the app. It will compile, but not execute even though it is defined as an application launch config.

image

Col-E: NVM im just dumb

xxDark:
> 
> 
> Its still a WIP branch, and I'm all ears.
> 
> I'd like to do the `::new` but retain module independence for `core` from `ui`. The presentation classes will be what populate and call all the UI code later, hence it being relocated to the ui module. Any ideas for a workaround?

Should not we keep launching process away from `core` mdule?
We can leave presentation classes in it, however, launching logic may be present in `ui` module.

Col-E:
> Should not we keep launching process away from core module?
> We can leave presentation classes in it, however, launching logic may be present in ui/cli modules.

Yeah, we can move launching to the UI module. 

However for presentation classes, since those will eventually directly reference UI logic, I want those to stay in the UI module.

xxDark:
> 
> 
> > Should not we keep launching process away from core module?
> > We can leave presentation classes in it, however, launching logic may be present in ui/cli modules.
> 
> Yeah, we can move launching to the UI module.
> 
> However for presentation classes, since those will eventually directly reference UI logic, I want those to stay in the UI module.

I  don't think that headless presentation really belongs to `ui` module, that's why I suggested to move it to `core` module.

Col-E:
The headless presentation is still UI driven, just the UI is a CLI.

The empty presentation is ok for the core since there is no UI associated with it.

Col-E:
Merged locally with some rebasing

PR #323, opened 2020-12-29
Implement common content sources

Col-E:
Work in progress for resource content sources.

Will need to create test cases to validate things function before merging.


recaf-core/src/main/java/me/coley/recaf/workspace/resource/source/MavenContentSource.java
```diff
@@ -0,0 +1,102 @@
...
+import static java.io.File.separatorChar;
+import static java.lang.String.format;
+
+/**
+ * Origin location information for maven artifacts.
+ *
+ * @author Matt Coley
+ */
+public class MavenContentSource extends ContentSource {
+	private static final String CENTRAL_REPO = "https://repo1.maven.org/maven2";
```
xxDark, line 23: I think it's a good idea to make this url customizeable.

Col-E: It can now be passed via the constructor

PR #337, opened 2021-01-12
Popups save location + basic YAML support

Strikeless:
## What's new

* Popups will now appear at the location of a previous popup if one exists
* Very basic YAML support/highlighting

## What's fixed

* N/A

### DIFF
--- a/src/main/java/me/coley/recaf/ui/controls/popup/DragPopup.java
+++ b/src/main/java/me/coley/recaf/ui/controls/popup/DragPopup.java
@@ -15,6 +15,8 @@ import me.coley.recaf.util.ThreadUtil;
  * @author Matt
  */
 public class DragPopup {
+	private static double lastX = -1D, lastY = -1D;
+
 	private final Popup pop = new Popup();
 
 	/**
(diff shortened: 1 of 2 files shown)


src/main/java/me/coley/recaf/ui/controls/popup/DragPopup.java
```diff
@@ -83,13 +86,18 @@ public void show(Node parent, double x, double y) {
...
 		ThreadUtil.runJfxDelayed(100, () -> {
 			pop.setOnShown(e -> {
-				Bounds bounds = parent.localToScreen(parent.getBoundsInParent());
-				double parentWidth = bounds.getMaxX() - bounds.getMinX();
-				double parentHeight = bounds.getMaxY() - bounds.getMinY();
-				double x = bounds.getMinX() + (parentWidth / 2);
-				double y = bounds.getMinY() + (parentHeight / 2);
-				pop.setX(x - pop.getWidth() / 2);
-				pop.setY(y - pop.getHeight() / 2);
+				if(lastX == -1D || lastY == -1D) {
```
Col-E, line 89: Just in case, it may be more safe to do `lastX < 0 || lastY < 0`

src/main/java/me/coley/recaf/ui/controls/popup/DragPopup.java
```diff
@@ -15,6 +15,9 @@
  * @author Matt
  */
 public class DragPopup {
+	private static double lastX = -1D;
+	private static double lastY = -1D;
```
Col-E, line 19: Is there a reason why we would want to retain the position and show new popups at the same location?

If a popup is originating from a `Node` that does not match this position _(Or if the user moves the window on the screen)_ then wouldn't that be confusing? I'm not sure I understand why this would be beneficial.

Strikeless: I was thinking that by making popups store their location it could make the experience a bit better when for example continuously getting decompiler errors on your way because of some well obfuscated code while only trying to look at something specific, but now that you brought up the other side of things I do agree it could get confusing and afterall there surely is a better solution to something like this.

PR #339, opened 2021-01-16
Javac target fix

sverkera:
## What's fixed

* I noticed that comparation of class version and max supported target version was wrong, resulting in that version of the target class version would always become the highest version supported by javac. In my case the original class was v7 but when I edited it would be compiled as v8 which was not correct and caused an issue as the resulting jar was to be run by a version 1.7 jvm.
*  Small fix for default of max target version, the log message and the returned version were not consistent.


Col-E:
Will merge in the next push 👍

PR #343, opened 2021-01-24
Support for mac key map.

TimmyOVO:
## What's new
Support for mac key map.

* Summary of additions
Instead of using ctrl+key shortcuts, osx is using ⌘(Command)+key shortcuts.
In Recaf's code viewer,default key binding for copy and paste is  ⌘(Command)+C and ⌘(Command)+V , I need use both ⌘(Command)+key and Ctrl+Key to get work done without supporting mac's keymap .

## What's fixed

NONE

* Summary of bugs fixed

### DIFF
--- a/src/main/java/me/coley/recaf/config/ConfKeybinding.java
+++ b/src/main/java/me/coley/recaf/config/ConfKeybinding.java
@@ -1,13 +1,5 @@
 package me.coley.recaf.config;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.Set;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
 import javafx.scene.Scene;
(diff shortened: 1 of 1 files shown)


Col-E:
Is it possible to update [`KeybindingField`](src/main/java/me/coley/recaf/ui/controls/KeybindField.java) to support ⌘?

I don't have a Mac but I recall somebody mentioned it doesn't work. 

As for changing the default binds, I don't have an issue with it, but am leaning on opposing it. I would rather have everyone use the same keybinds by default and allow them to make changes on their own. For instance, each time a bind is mentioned having to say _"Its this XYZ unless you're on mac the nits ABC"_ is a bit tedious. If a user changes they keybinds they're probably already aware they have done so.

TimmyOVO:
Because some of the key binding are using command instead using control( Copy and paste etc.) ,this is bit of annoying
, those key bindings changed their control key to command depends on what system you are using right now, i think other keybindings should follow too.

Issue #344, opened 2021-01-26
StringIndexOutOfBoundsException in CfrDecompiler.clean


shoffmeister:
Upon decompiling a specific class in Recaf 2.17.3 with CFR, the Recaf CFR decompiler _integration_ (not the decompiler itself) runs into "java.lang.StringIndexOutOfBoundsException"

**To Reproduce**

* Download the JAR for https://mvnrepository.com/artifact/com.google.code.gson/gson/2.2.4 (https://repo1.maven.org/maven2/com/google/code/gson/gson/2.2.4/gson-2.2.4.jar)
* File -> Load -> gson-2.2.4.jar
* Decompile class "com/google/gson/internal/$Gson$Types$GenericArrayTypeImpl"

//exp: decompiles successfully
//act: java.lang.StringIndexOutOfBoundsException

**Exception**

From console output:
```
08:38:40.837 [pool-6-thread-3] ERROR: Class failed to decompiled: com/google/gson/internal/$Gson$Types$GenericArrayTypeImpl
java.lang.StringIndexOutOfBoundsException: String index out of range: -1
        at java.lang.String.substring(String.java:1967)
        at me.coley.recaf.decompile.cfr.CfrDecompiler.clean(CfrDecompiler.java:93)
        at me.coley.recaf.decompile.cfr.CfrDecompiler.decompile(CfrDecompiler.java:64)
        at me.coley.plugin.decompile.DecompilePanel.lambda$startDecompile$2(DecompilePanel.java:78)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```
i.e. the crash is not in CFR (which I validated to decompile), but in the integration of CFR at https://github.com/Col-E/Recaf/blob/master/src/main/java/me/coley/recaf/decompile/cfr/CfrDecompiler.java#L93

shoffmeister:
This is not limited to this specific .jar, but apparently a more generic problem.

[...]

PR #346, opened 2021-01-26
Be more resilient when trying to fixup CFR class names

shoffmeister:
CFR generates class names which will not (re)compile correctly. Recaf attempts to fix up those generated class names, to get to compilable source code. Alas, due to the way CFR generates those c

This change here comes on top of https://github.com/Col-E/Recaf/issues/344, where the associated changes applied still may result in StringIndexOutOfBoundsException (see comment inside code).

This PR makes the relevant code more *resilient* against decompiled source code / CFR behaviour; it prefers to have "some" CFR-generated code (possibly with a comment at the top indicating an oops) over getting a StringIndexOutOfBoundsException (and no output from CFR).


Col-E:
Ah, nice catch! Will include in the next push.

Col-E:
Messed up some rebasing so github doesn't recognize that its merged properly, but the changes are included!

8e0b34d6343f7a01835ddb5788057bcdc223a1d0
619853c2464d90fd69b0d405f260d177511bf235

PR #347, opened 2021-01-28
3X: Begin laying out UI module skeleton

Col-E:
## What's new

* A rough draft of how the UI flows from `main` to opening the main window.


Col-E:
Current state of the UI:

image

Future PR's will revolve around adding actual content to the main region and additional features adjacent to the logging tab

PR #352, opened 2021-02-13
feat: Diversified font selection & added 'Swap View' keybind

codingsocks:
## What's new

* You can now select separate fonts for the UI and the monospaced / editor fonts, as well as the sizes
* You can now swap views using a keybind (defaulted to CTRL+Q)


yapht:
poggers. liked.

PR #360, opened 2021-03-01
Escape abritrary strings in assembler instructions

x4e:
## What's new

Some instructions that have the ability to print arbitrary strings will have the printed contents escaped, so some non alphanumeric characters will be replaced with escape sequences.
These escape sequences will be decoded upon compilation.

## What's fixed

Previously some class names, especially those with unicode characters and new line characters, would be printed by the disassembler output, resulting in an invalid input for the assembler.

[example.zip](https://github.com/Col-E/Recaf/files/6058472/example.zip)

There are still many places that I can see where the assembler can be exploited, however this should fix one common place.
This was especially noticeable in field instructions accessing fields in classes with obfuscated names.

I have run some of my own tests as well as the official tests and see no errors, so it should be fine.

### DIFF
--- a/src/main/java/me/coley/recaf/parse/bytecode/Disassembler.java
+++ b/src/main/java/me/coley/recaf/parse/bytecode/Disassembler.java
@@ -279,15 +279,22 @@ public class Disassembler {
 		// Sometimes its not though.
 		Type type = insn.desc.contains(";") ?
 				Type.getType(insn.desc) : Type.getObjectType(insn.desc);
-		line.append(' ').append(type.getInternalName());
+		String name = EscapeUtil.escape(type.getInternalName());
+		line.append(' ').append(name);
 	}
 
 	private void visitFieldInsn(StringBuilder line, FieldInsnNode insn) {
-		line.append(' ').append(insn.owner).append('.').append(insn.name).append(' ').append(insn.desc);
+		String owner = EscapeUtil.escape(insn.owner);
+		String name = EscapeUtil.escape(insn.name);
+		String desc = EscapeUtil.escape(insn.desc);
+		line.append(' ').append(owner).append('.').append(name).append(' ').append(desc);
 	}
 
 	private void visitMethodInsn(StringBuilder line, MethodInsnNode insn) {
-		line.append(' ').append(insn.owner).append('.').append(insn.name).append(insn.desc);
+		String owner = EscapeUtil.escapeCommon(insn.owner);
+		String name = EscapeUtil.escapeCommon(insn.name);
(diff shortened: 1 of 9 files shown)


Crystallinqq:
this is good

src/main/java/me/coley/recaf/parse/bytecode/Disassembler.java
```diff
@@ -307,7 +314,8 @@ private void visitLdcInsn(StringBuilder line, LdcInsnNode insn) {
 		line.append(' ');
 		if(insn.cst instanceof String) {
 			String str = insn.cst.toString();
-			str = EscapeUtil.escapeCommon(str);
+			// TODO: support for escaping "
+			str = EscapeUtil.escape(str);
```
Col-E, line 318: Because our parser is a hacky pile of substringing mess, I don't think we actually need to escape `"` for now.

src/main/java/me/coley/recaf/parse/bytecode/ast/NameAST.java
```diff
@@ -27,6 +29,13 @@ public NameAST(int line, int start, String name) {
 	public String getName() {
 		return name;
 	}
+	
+	/**
+	 * @return Name without escapes.
+	 */
+	public String getUnescapedName() {
+		return EscapeUtil.unescape(name);
+	}
```
Col-E, line 38: Something about having two name getters feels wrong. I get that this works, but its just a gut feeling.

src/main/java/me/coley/recaf/parse/bytecode/ast/TypeAST.java
```diff
@@ -28,6 +30,13 @@ public String getType() {
 		return type;
 	}
 
+	/**
+	 * @return Internal type without escapes.
+	 */
+	public String getUnescapedType() {
+		return EscapeUtil.unescape(type);
+	}
```
Col-E, line 38: Same comment as for the duplicate name getters above.

src/main/java/me/coley/recaf/parse/bytecode/ast/NameAST.java
x4e: I wanted to maintain compatibility with other parts of the code that may use names without escaping

Crystallinqq:
Please merge this.

Col-E:
I will, I merge locally so I can adjust version constants and pull in multiple items at once for the automated release.

Crystallinqq:
swagalicious.

x4e:
Thanks for being so swag

Col-E:
Merged locally, github doesn't sync if you rebase something even if you don't directly touch the commits.

PR #368, opened 2021-04-03
feat: tsrg mappings

1fxe:
## What's new

- Added support for tsrg mappings.
- Added tsrg test

### DIFF
--- a/src/main/java/me/coley/recaf/mapping/TSrgMappings.java
+++ b/src/main/java/me/coley/recaf/mapping/TSrgMappings.java
@@ -38,6 +38,9 @@ public class TSrgMappings extends FileMappings {
             String[] args = lineStr.trim().split(" ");
             try {
                 // Fields and Methods start with a tab
+                // class/Name new/Name
+                //      fieldName newFieldName
+                //      methodName methodDesc newMethodName
                 if (!lineStr.startsWith("\t")) {
                     obfOwner = args[0];
                     String renamedClass = args[1];
--- a/src/main/java/me/coley/recaf/mapping/TSrgMappings.java
+++ b/src/main/java/me/coley/recaf/mapping/TSrgMappings.java
@@ -9,7 +9,7 @@ import java.util.HashMap;
 import java.util.Map;
 
 /**
- * SRG mappings file implementation.
+ * TSRG mappings file implementation.
  *
  * @author Filip (1fxe)
  */
--- a/src/main/java/me/coley/recaf/mapping/TSrgMappings.java
+++ b/src/main/java/me/coley/recaf/mapping/TSrgMappings.java
@@ -29,8 +29,8 @@ public class TSrgMappings extends FileMappings {
 
     @Override
     protected Map<String, String> parse(String text) {
-        Map<String, String> map = new HashMap<>();
         String[] lines = StringUtil.splitNewline(text);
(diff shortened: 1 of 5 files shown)


Col-E:
👍 will include the next time I push some commits

src/main/java/me/coley/recaf/mapping/TSrgMappings.java
```diff
@@ -0,0 +1,66 @@
...
+     * @param workspace Workspace to pull names from when using hierarchy lookups.
+     * @throws IOException Thrown if the file could not be read.
+     */
+    TSrgMappings(Path path, Workspace workspace) throws IOException {
+        super(path, workspace);
+    }
+
+    @Override
+    protected Map<String, String> parse(String text) {
+        Map<String, String> map = new HashMap<>();
```
x4e, line 32: There is nearly always one map entry per line of text. Perhaps it would be good to initialise the HashMap after the `StringUtil.splitNewline` and pass it `lines.length`?

src/main/java/me/coley/recaf/mapping/TSrgMappings.java
```diff
@@ -0,0 +1,66 @@
...
+                        String renamedName = args[1];
+                        map.put(obfOwner + "." + obfName, renamedName);
+                    } else if (args.length == 3) { // Method
+                        String obfName = args[0];
+                        String obfDesc = args[1];
+                        String renamedName = args[2];
+                        map.put(obfOwner + "." + obfName + obfDesc, renamedName);
+                    }
+                }
+            } catch (IndexOutOfBoundsException ex) {
```
x4e, line 60: Could this not just be `Throwable` so that any error that may occur gets rethrown with the nice error message?

src/main/java/me/coley/recaf/mapping/TSrgMappings.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * TSRG mappings file implementation.
+ *
+ * @author Filip (1fxe)
+ */
+public class TSrgMappings extends FileMappings {
+    private static final String FAIL = "Invalid TSrg mappings, ";
```
x4e, line 17: Does this really need to be a class level constant? And if so, why not have `"failed parsing line "` as a constant too? Seems inconsistent.

1fxe: I wrote the constant to be consistent with the other mapping files. For example https://github.com/Col-E/Recaf/blob/419925d996fe33d3caa94a864070af97c1a2d625/src/main/java/me/coley/recaf/mapping/SrgMappings.java#L19

https://github.com/Col-E/Recaf/blob/419925d996fe33d3caa94a864070af97c1a2d625/src/main/java/me/coley/recaf/mapping/TinyV1Mappings.java#L19

x4e: Ah ok, looks good then

src/main/java/me/coley/recaf/mapping/TSrgMappings.java
1fxe: For consistency with other mapping classes I will leave it, maybe open an issue to address this with other mappings?

PR #376, opened 2021-04-25
General keybind fixes

codingsocks:
## What's fixed

* Fixed bad swap view default bind
* Fixed close window keybind activating if you edit it and set it to itself
* Fixed keybind fields keeping focus after a keybind is set


codingsocks:
The close window keybind thing should probably occur on all keybinds, but I didn't add that in this PR.

PR #381, opened 2021-05-16
Features

1fxe:
## What's new

- Remap command will now remap the manifest main-class attribute
-  class/field/member modifier groups now have a tooltip with their access flags

## What's fixed

- Fixed issue where context menu would change color on hover


1fxe (commented):
Need to write tests for this later

src/main/java/me/coley/recaf/util/UiUtil.java
```diff
@@ -176,9 +181,24 @@ else if(AccessFlag.isAbstract(access))
...
-		return g;
+        createAccessToolTips(g, AccessFlag.Type.METHOD, access);
+        return g;
 	}
 
+    private static void createAccessToolTips(Node node, AccessFlag.Type type, int access) {
+        Set<String> accessFlags = AccessFlag.getApplicableFlags(type, access).stream()
+				.map(AccessFlag::getName).collect(Collectors.toSet());
+        Tooltip tooltip = new Tooltip(String.join(", ", accessFlags));
+        node.setOnMouseEntered(event -> {
```
1fxe, line 192: Add a config option? Enabled by default maybe

src/main/java/me/coley/recaf/command/impl/Remap.java
```diff
@@ -48,8 +54,30 @@ public Void call() throws Exception {
...
+				String mainClass = attr.getValue("Main-Class").replaceAll("\\.", "/");
+				if (mapped.containsKey(mainClass)) {
+					info("Found Main-Class attribute!");
+					attr.putValue("Main-Class", new ClassReader(mapped.get(mainClass))
+							.getClassName().replaceAll("/", "\\."));
+					ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+					manifest.write(outputStream);
+					primary.getFiles().put("META-INF/MANIFEST.MF", outputStream.toByteArray());
+					outputStream.close();
+					info("Remapped manifest!");
```
Col-E, line 76: Logging level can probably be dropped to debug. And I don't think they need to end with `!` either.

src/main/java/me/coley/recaf/util/UiUtil.java
```diff
@@ -176,9 +182,24 @@ else if(AccessFlag.isAbstract(access))
...
+				.map(AccessFlag::getName).collect(Collectors.toSet());
+		Tooltip tooltip = new Tooltip(String.join(", ", accessFlags));
+		node.setOnMouseEntered(event -> {
+			if (!tooltip.getText().isEmpty() && Recaf.getController().config().display().accessFlagsTooltip) {
+				tooltip.show(node, event.getScreenX(), event.getScreenY() + 15);
+			}
+		});
+		node.setOnMouseExited(event -> tooltip.hide());
+		// This has a weird delay
+		// Tooltip.install(node, tooltip);
```
Col-E, line 200: I'm pretty sure the idea is that normally tooltips show after hovering the mouse for a short time. It's intentionally designed that way.  Though I do understand why making this show up instantly makes more sense 👍

Col-E: I realized some components already have access flag tooltips, such as in `ClassNodeEditorPane` which makes two pop-ups show up when hovering. I'll fix this in merge.

PR #386, opened 2021-05-24
3X: Base code editor

Col-E:
**Differences from 2.X**

 - Design changes
     - Feature support is not baked into the code editor class's generic arguments. 
 - Optimizations
     - Only restyles the lines necessary for each edit _(Adding a comment `//` will no longer trigger restyling the whole file)_
     - Error indicators no longer regenerate for each edit _(Because they now know how to update positions when the text updates)_
 - New features
     - Has `{}` / `[]` / `<>` / `()` bracket highlighting

**Screenshot**

image


Col-E:
There's really only one remaining issue with the optimized _"only restyle what is needed"_ approach:

https://user-images.githubusercontent.com/21371686/119303024-0e0cb580-bc33-11eb-8c97-988b9c6dad0d.mp4

Since the system is forward-looking if you put the opening `/*` last it sees the pattern. If you put the `*/` last, it looks at the line's contents and doesn't see the pattern beginning a couple lines up. Not sure how to go about fixing this atm...

recaf-ui/src/main/java/me/coley/recaf/ui/test/FooMain.java
```diff
@@ -0,0 +1,65 @@
...
+import org.fxmisc.flowless.VirtualizedScrollPane;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * Testing.
+ */
+public class FooMain extends Application {
```
Col-E, line 17: Ooops, didn't mean to commit this file

Col-E:
And now we have code folding :)

https://user-images.githubusercontent.com/21371686/120269724-61f04d80-c276-11eb-967b-296825895fc3.mp4

PR #393, opened 2021-06-17
Update file detection/saving

### LINKED ISSUE #391: Saving to a network location takes a very long time
When saving to a Windows network shared drive, it takes a very long time (a 2K file took at least 5 minutes). Saving locally is almost instantly

xxDark:
## What's new

Detect an input based on file headers
Detect class files by it's header
Optimize directory pruning
Proper (?) dock name for MacOS users

## What's fixed

Fixed #391 
Fixed an error when user attempts to load invalid input

### DIFF
--- a/src/main/java/me/coley/recaf/command/impl/Export.java
+++ b/src/main/java/me/coley/recaf/command/impl/Export.java
@@ -13,6 +13,7 @@ import java.io.*;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
 import java.util.*;
 import java.util.concurrent.Callable;
 import java.util.jar.JarEntry;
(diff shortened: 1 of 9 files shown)


src/main/java/me/coley/recaf/util/IOUtil.java
```diff
@@ -395,4 +458,36 @@ public static Path toPath(File file) {
...
+    }
+
+    /**
+     * @param bytes byte array to check.
+     * @return {@code true} if byte array represents
+     * a class file header.
+     */
+    public static boolean isClassHeader(byte[] bytes) {
+        return bytes[0] == (byte) 0xCA && bytes[1] == (byte) 0xFE
+                && bytes[2] == (byte) 0xBA && bytes[3] == (byte) 0xBE;
```
Col-E, line 479: Another variant of this exists as `ClassUtil#isClass(byte[] data)`, do we want to remove that or use it?

src/main/java/me/coley/recaf/util/IOUtil.java
```diff
@@ -395,4 +458,36 @@ public static Path toPath(File file) {
...
+    /**
+     * @param bytes byte array to check.
+     * @return {@code true} if byte array represents
+     * jar file signature.
+     */
+    public static boolean isJarSignature(byte[] bytes) {
+        if (bytes.length < 2) {
+            return false;
+        }
+        return bytes[0] == -2 && bytes[1] == -54;
```
Col-E, line 491: The other methods don't have length checks for the input byte array, do we want to add this check to the other methods to prevent possible `IndexOutOfBoundsException`?

src/main/java/me/coley/recaf/command/impl/Export.java
```diff
@@ -162,6 +165,8 @@ public static void writeArchive(File output, Map<String, byte[]> content) throws
 				jos.closeEntry();
 			}
 		}
+		Files.write(output.toPath(), bytes.toByteArray(),
+				StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING);
```
Col-E, line 169: Looks like the options here do not need to be specified: https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#write(java.nio.file.Path,%20byte[],%20java.nio.file.OpenOption...)

> If no options are present then this method works as if the CREATE, TRUNCATE_EXISTING, and WRITE options are present.

Col-E (commented):
Few minor questions, but otherwise LGTM 👍

Col-E:
Tested exporting changes out on `rt.jar` _(60 mb)_ and it is noticeably faster. Compared memory and it does rise faster _but_ a single GC wipes all of the data away 👍 

Before:

image

After:

image

PR #410, opened 2021-10-17
Fix class reading with bad CRC

fiwka:
## What's fixed

* Fixed class reading with bad CRC


Col-E:
If you want an example I implemented this fix on the `dev3` branch while also retaining the existing behavior: a84d59d6af02f9421202d7938996f1f9b5075746

win32kbase:
No activity since October, closing this for now

PR #417, opened 2021-11-07
Allow zip files to be opened as jar files

Tigermouthbear:
### Allows zip files to be opened like jar files
Since I use Recaf to analyze malicious programs, I'd rather store the programs as .zip files so that they don't run when I click them. This PR just adds zip to the file extensions which are allowed to be opened with Recaf.


Col-E:
Drag and drop support already supports this, same situation for loading items from the recent menu. The only place where I can't load a renamed jar is from the `File > Load` prompt, but that's only because the file filter option doesn't have an `All files (*)` alternative.

Tigermouthbear:
oh, I guess I was using an older version lmao

Col-E:
👌

PR #434, opened 2021-12-17
Fix export directory option in PE explorer pane appearing when no export directory is available

win32kbase:
## What's fixed

* Fix export directory option in PE explorer pane appearing when no export directory is available


recaf-ui/src/main/java/me/coley/recaf/ui/pane/pe/PEExplorerPane.java
```diff
@@ -92,7 +93,14 @@ public void onUpdate(FileInfo newValue) {
 		LittleEndianReader reader = new LittleEndianReader(stream);
 		pe = ImagePeHeaders.read(reader);
 
-		// Reset section headers
+		// Remove the export directory option if the file doesn't have one
+		if (dummyRoot.getChildren().contains(itemExportDirectory)) {
+			if (pe.getCachedExports() == null) {
```
Col-E, line 98: Can be an && in a single statement

PR #461, opened 2022-01-27
Added german translation

Nowilltolife:
## What's new

Added german translation

## What's fixed

no bug fixed


win32kbase:
Thanks!

PR #464, opened 2022-01-28
System language defaulting 

Nowilltolife:
## What's new

- Added system language detection and made it load as default if no language is set
(format: language_COUNTRY, e.g: en_US, de_DE)
- Changed en.lang -> en_US.lang to fit detection format

## What's fixed

No bugs fixed


3jq (approved):
cool!

PR #492, opened 2022-02-25
Update dependencies and Maven wrapper

ThexXTURBOXx:
Similar PR will arrive shortly for 3.X

## What's new

* Updated dependencies
* Updated Maven wrapper

## What's fixed

* Resulting Compile errors fixed


src/main/java/me/coley/recaf/util/JavaParserUtil.java
```diff
@@ -506,7 +506,7 @@ private static String typeToDesc(Type type) {
 			ClassOrInterfaceType clsType = (ClassOrInterfaceType) type;
 			clsType.setTypeArguments(NO_TYPE_ARGS);
 			try {
-				key = toInternal(clsType.resolve().getTypeDeclaration().get());
+				key = toInternal(clsType.resolve().asTypeParameter());
```
Col-E, line 509: This will throw an exception when used:

```java
    default ResolvedTypeParameterDeclaration asTypeParameter() {
        throw new UnsupportedOperationException(String.format("%s is not a Type parameter", this));
    }
```

Did you mean to use `asReferenceType()`?

src/main/java/me/coley/recaf/util/IllegalBytecodePatcherUtil.java
```diff
@@ -1,6 +1,6 @@
 package me.coley.recaf.util;
 
-import me.coley.cafedude.ClassFile;
+import me.coley.cafedude.classfile.ClassFile;
```
Col-E, line 3: Part of the 1.8 to 1.9 transition requires adding `new IllegalStrippingTransformer(cf).transform();` to retain the same behavior of stripping invalid attributes.

Col-E:
Two comments aside, LGTM 👍

xxDark:
Noting that maven wrapper is not needed for 3.x, we use Gradle

ThexXTURBOXx:
@Col-E Thanks, will reply solve issues soon!

@xxDark Yes, I also updated the Gradle wrapper there already :) Thanks for the note though!

Col-E:
👍

PR #493, opened 2022-02-25
Update dependencies and Gradle wrapper

ThexXTURBOXx:
## What's new

* Update dependencies
* Update Gradle wrapper


recaf-core/src/main/java/me/coley/recaf/workspace/resource/ClassPatchingListener.java
```diff
@@ -1,7 +1,7 @@
 package me.coley.recaf.workspace.resource;
 
-import me.coley.cafedude.ClassFile;
 import me.coley.cafedude.InvalidClassException;
+import me.coley.cafedude.classfile.ClassFile;
```
Col-E, line 4: Same as on the 2.X PR, needs the transformer applied before the class bytes are written back

Col-E:
Again, 👍

PR #505, opened 2022-03-13
Use Gradle Java Toolchains instead of having requirement on Java version required to run gradle.

0-x-2-2:
https://docs.gradle.org/current/userguide/toolchains.html

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,19 +9,21 @@ allprojects {
     version '3.0.0-SNAPSHOT'
 }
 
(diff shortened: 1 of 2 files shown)


0-x-2-2:
Haven't tested on another machine, so hopefully, you have CI testing...

Col-E:
Works on my machine:tm: _(And doesn't affect CI running on 17)_

0-x-2-2:
lgtm

PR #507, opened 2022-03-15
Load plugin translation files

DasDirt:
## What's new

* The translation file of plugins will now be loaded

## What's fixed

* none


Col-E:
I recognize that there is no clear way to get plugins to load the _current_ language without some hacky logic, so for now this is good.

The 3X model does remedy this, so once the plugin system is integrated over there this feature can be more useful.

PR #508, opened 2022-03-17
Feature/improved docking api

Col-E:
## What's new

* Replaced the docking system of `DockingRootPane` with something much cleaner.


Col-E:
Looks like there's a bug with class renaming. The old docking system wasn't calling `cleanup()` on tabs being closed for renaming purposes. The new docking system has better tracking support, so it does call `cleanup()`. Since we re-use the UI control from the old tab and `cleanup()` closes the thread-pools it crashes...

PR #510, opened 2022-03-23
Configurable language associations

yapht:
## What's new

* Allow users to change associations between file extensions and languages


Col-E:
A preview:

https://user-images.githubusercontent.com/21371686/159902014-fa486b8f-6ae2-4f67-af45-c10a09fb6ff8.mp4

Issue #513, opened 2022-03-28
Escaped unicode characters are decoded wrongly


L3g7:
When having an escaped unicode character (`"\\u0041"`), Recaf decodes the unicode character (`"\A"`), resulting in a lexical error.

To reproduce:
1. compile following code
2. open with recaf
```java
class Example {
    String example = "\\u0041";
}
```
![grafik](https://user-images.githubusercontent.com/62625188/160480184-bb37db50-6268-40b5-8ee9-482275104060.png)

win32kbase:
Thanks. Think it is fixed in above PR.

Image:
![image](https://user-images.githubusercontent.com/6201484/160492645-904d5ee6-7839-4fe8-ac87-b629839b40f3.png)

Col-E:
I actually think this is something that should be reported and addressed upstream in CFR. That would be a proper solution rather than us _patching_ the output.

win32kbase:
This is now fixed in both 2X and 3X

PR #514, opened 2022-03-28
Fix double backslash before escaped unicode character causing lexing error

win32kbase:
Fix double backslash before escaped unicode character causing lexing error

## What's fixed

* Fixes https://github.com/Col-E/Recaf/issues/513


Col-E:
Does this affect disassembler output in such a way that we would need to change how unescapes work?

So for instance, the same char escape as in the linked issue, but in the assembler rather than the decompiler view?

Col-E:
Did some testing, seems to not have any regressions. Merging 👍

Issue #446, opened 2022-01-07
encoding issues


Tukali:
If there is a Chinese in the file, it will be garbled
![image](https://user-images.githubusercontent.com/45891208/148531717-5e436f33-9f38-480c-8c86-a1f7386cdcd3.png)

3jq:
Probably recaf just gotta use utf-16 for decompiled text.

xtherk:
use `java -jar -Dfile.encoding=UTF-8 recaf.jar` to run

xtherk:
> If there is a Chinese in the file, it will be garbled ![image](https://user-images.githubusercontent.com/45891208/148531717-5e436f33-9f38-480c-8c86-a1f7386cdcd3.png)

### Reason
Your Windows system defaults to GBK, you double-click `recaf.jar` or use `java -jar recaf.jar` Run, the default characters inside the Java are GBK, the Recaf processing text is use ` new String(Byte[])`, this when the default encoding is gbk, it will cause garbled.

Toshimichi0915:
we can close this issue?

win32kbase:
> we can close this issue?

Yep, thank you for your PR.

V-KOOL:
Have you solved this Chinese garbled code problem?

PR #524, opened 2022-04-20
Use utf-8 instead of default charset in FileViewport text mode

Toshimichi0915:
## What's fixed
It's reasonable to show text in UTF-8 instead of system default charset as most java developers use UTF-8.
This PR fixes encoding issue in text mode.
[sample.zip](https://github.com/Col-E/Recaf/files/8519355/sample.zip)


Toshimichi0915:
Fixes #446

Toshimichi0915:
image
I just checked Japanese characters in my jar file and everything seemed fine in every decompiler.
I guess decompilers constructs String in UTF-8 internally

Col-E:
Ok, looks good to me then

PR #526, opened 2022-04-23
3X: Android decompilation support [WIP]

Col-E:
## What's new

* Decompilation support for dex classes

## Considerations

There is a lot of refactoring of the UI classes inside this.
So far, the implementation idea is to map DexLib --> JadX so we can use JadX to decompile dex classes. JadX is mostly designed with Android Dalvik in mind, so the mapping process _is_ simpler than a complete mapping from Dalvik to Java. However, if we are putting in the effort for this mapping I have some questions:

- Should we put in an extra mile to do a complete DexLib -> Java so _any_ decompiler can be used?

    - What are our options for this that already exist?
         - [Dex2Jar](https://github.com/pxb1988/dex2jar) - designed as a CLI tool, the code behind it is messy
         - [Dalvikgate](https://github.com/GraxCode/dalvikgate) - abandoned
    - Would it be too complex to make our own mapper?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -47,7 +47,7 @@ subprojects {
         def asmVersion = '9.2'
         def dexVersion = '2.5.2'
         def dudeVersion = '1.10.0'
-        def jadxVersion = '1.3.3'
+        def jadxVersion = '1.3.4'
         def jpVersion = '3.24.0'
         def junitVersion = '5.8.2'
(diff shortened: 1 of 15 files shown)


Col-E:
Closing, an alternative approach using https://github.com/Col-E/dex-translator is in place on a separate branch

PR #539, opened 2022-05-22
Assembler context menus for instruction

Nowilltolife:
## What's new

- Added context menus for instructions

## What's fixed

- Bugs 🎲 🤔 
- Annotations not showing up on field
- Signature not showing up at all
- Lables not wrapped

### DIFF
--- a/recaf-assembler/src/main/java/me/coley/recaf/assemble/ast/BaseElement.java
+++ b/recaf-assembler/src/main/java/me/coley/recaf/assemble/ast/BaseElement.java
@@ -149,6 +149,15 @@ public abstract class BaseElement implements Element {
 
 	}
 
+	public Element getChildAt(int position) {
+		for (Element child : getChildren()) {
+			if (child.getStart() <= position && child.getStop() >= position) {
+				return child;
+			}
+		}
+		return null;
+	}
+
 	@Override
 	public String toString() {
 		return print();
--- a/recaf-assembler/src/main/java/me/coley/recaf/assemble/ast/BaseElement.java
+++ b/recaf-assembler/src/main/java/me/coley/recaf/assemble/ast/BaseElement.java
@@ -148,11 +148,29 @@ public abstract class BaseElement implements Element {
 		return closestElement;
 
 	}
-
 	public Element getChildAt(int position) {
-		for (Element child : getChildren()) {
-			if (child.getStart() <= position && child.getStop() >= position) {
-				return child;
+		Element[] elements = getChildren().toArray(new Element[0]);
+		if (elements.length == 0) { // base case
+			return null;
+		}
+		if (elements.length == 1) { // base case
+			return elements[0];
+		}
+
+		// perform binary search
+		int start = 0;
+		int end = elements.length - 1;
+
+		while (start <= end) {
+			int mid = (start + end) / 2;
+			if (position < elements[mid].getStart()) {
+				end = mid - 1;
+			}
+			else if (position > elements[mid].getStop()) {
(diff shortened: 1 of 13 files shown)


recaf-assembler/src/main/java/me/coley/recaf/assemble/ast/BaseElement.java
```diff
@@ -149,6 +149,15 @@ public Element getChildAt(int line, int column) {
 
 	}
 
+	public Element getChildAt(int position) {
+		for (Element child : getChildren()) {
+			if (child.getStart() <= position && child.getStop() >= position) {
```
Col-E, line 154: Incredibly nit-picky, but I think `pos >= start && pos <= stop` is more immediately legible than `start <= pos && stop >= pos`.

Supporting evidence:
 - My sleep deprived brain was confused at first glance here

recaf-assembler/src/main/java/me/coley/recaf/assemble/ast/BaseElement.java
```diff
@@ -149,6 +149,15 @@ public Element getChildAt(int line, int column) {
...
+				return child;
+			}
+		}
```
Col-E, line 157: Optimization: Can use binary search or predictive search on `Element`'s positions to skip over most elements.

recaf-assembler/src/main/java/me/coley/recaf/assemble/pipeline/AssemblerPipeline.java
```diff
@@ -259,6 +260,10 @@ public Unit getUnit() {
 		return unit;
 	}
 
+	public List<Group> getLatestAst() {
```
Col-E, line 263: Will look over usage later, but having this being _"latest ast"_ may be confusing when comparing to _"unit"_ which is also occasionally referred to as the AST. Could make the naming convention of Jasm AST more specific to differentiate from Recaf AST?

recaf-ui/src/main/java/me/coley/recaf/ui/control/code/bytecode/AssemblerArea.java
```diff
@@ -43,467 +49,637 @@
...
+import static me.coley.recaf.ui.util.Menus.action;
+import static me.darknet.assembler.parser.Group.GroupType;
 
 /**
  * Text editing portion of the assembler UI.
  *
  * @author Matt Coley
  */
 public class AssemblerArea extends SyntaxArea implements MemberEditor,
-		AstValidationListener, BytecodeValidationListener, ParserFailureListener, BytecodeFailureListener {
```
Col-E, line 59: Your auto-format is set to use spaces, where the rest of Recaf's source is formatted with tabs. The diff here doesn't need to be this large.

Col-E (changes requested):
Some questions to address.

Minor: Some missing docs and a few formatting things. I can fix those myself though.

recaf-assembler/src/main/java/me/coley/recaf/assemble/pipeline/AssemblerPipeline.java
Nowilltolife: i thought about Parser AST

recaf-ui/src/main/java/me/coley/recaf/ui/control/code/bytecode/AssemblerArea.java
Nowilltolife: yes i have noticed

PR #560, opened 2022-07-12
3X: Optimized bytecode backed context resolving

Col-E:
The performance is notably better than the existing Javassist based bytecode backed symbol solver types... but JavaParser's source is incredibly messy and there is a lot of ugly static interdependence on utility classes. ~~That sort of setup cannot be easily extensible, which makes it impossible as is to support all forms of resolution that the standard bundle does~~ Thankfully we can successfully work around it!


Col-E:
Holy shit the amount of debugging time I sunk into 2f4487785131ed13e485fb597f056934083f6e9b is painful to think about....

PR #576, opened 2022-07-31
Add some QoL in class members tab

Amejonah1200:
## What's new
- [x] Search field
- [x] Member type filter
- [x] Visibility filter
- [x] Alphabetical sort
- [x] Visibility sort 
- [x] Nested/inner class (?)
- [x] Outer class breadcrumbs
- [x] Finishing touches
  - [x] Add copyright notice from Jetbrains
  - [x] documentation
  - [x] "Oracle's Java Virtual Machine implementation does not check the consistency of an InnerClasses attribute against a class file representing a class or interface referenced by the attribute. "
    seems like we need to check cyclic and strange phenomenon caused by obfuscators, just to be safe


recaf-ui/src/main/java/me/coley/recaf/ui/pane/OutlinePane.java
```diff
@@ -31,10 +31,9 @@
  * @author Matt Coley
  */
 public class OutlinePane extends BorderPane implements ClassRepresentation {
-	private static final SimpleBooleanProperty showTypes = new SimpleBooleanProperty();
-	private static final SimpleBooleanProperty showSynthetics = new SimpleBooleanProperty(true);
-	private final OutlineTree tree = new OutlineTree();
-	private final ClassRepresentation parent;
+	public static final SimpleBooleanProperty showTypes = new SimpleBooleanProperty();
+	public static final SimpleBooleanProperty showSynthetics = new SimpleBooleanProperty(true);
```
Col-E, line 35: `public static` makes it feel icky across classes, but I don't really refactoring to do otherwise as worth the effort.

recaf-ui/src/main/java/me/coley/recaf/ui/control/tree/OutlineTree.java
```diff
@@ -0,0 +1,144 @@
...
+
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
+import static me.coley.recaf.ui.pane.OutlinePane.showSynthetics;
+import static me.coley.recaf.ui.pane.OutlinePane.showTypes;
+
+/**
+ * Tree that represents the {@link MemberInfo} of a {@link CommonClassInfo}.
+ */
```
Col-E, line 27: Can add back `@author` from the class this was pulled from now that its in its own file.

recaf-ui/src/main/java/me/coley/recaf/ui/control/tree/OutlineTreeWrapper.java
```diff
@@ -0,0 +1,63 @@
...
+
+public class OutlineTreeWrapper extends OutlineTree{
+
+	private final SimpleBooleanProperty caseSensitivity = new SimpleBooleanProperty();
+
+	private final OutlineTree tree;
+
+	private final ObservableStringValue filter;
+
+	public OutlineTreeWrapper(ClassRepresentation parent, ObservableStringValue filter) {
```
Col-E, line 23: Documentation once finalized.

recaf-core/src/main/java/me/coley/recaf/code/ClassInfo.java
```diff
@@ -172,6 +181,12 @@ public MethodVisitor visitMethod(int access, String name, String descriptor, Str
 				methods.add(new MethodInfo(className, name, descriptor, sig, access, exceptions));
 				return null;
 			}
+
+			@Override
+			public void visitInnerClass(String name, String outerName, String innerName, int access) {
+				if(Objects.equals(name, className)) return;
```
NoJokeFNA, line 187:
```suggestion
				if (Objects.equals(name, className)) return;
```

recaf-ui/src/main/java/me/coley/recaf/ui/control/config/ConfigBoolean.java
```diff
@@ -23,7 +26,15 @@ public class ConfigBoolean extends CheckBox {
 	 */
 	public ConfigBoolean(ConfigContainer instance, Field field, ObservableValue<String> text) {
 		textProperty().bind(text);
-		setSelected(ReflectUtil.quietGet(instance, field));
-		selectedProperty().addListener((observable, old, current) -> ReflectUtil.quietSet(instance, field, current));
+		if (ObservableBooleanValue.class.isAssignableFrom(field.getType())) {
+			var value = (ObservableBooleanValue) ReflectUtil.quietGet(instance, field);
+			if (value instanceof BooleanProperty)
```
NoJokeFNA, line 31:
```suggestion
			if (value instanceof BooleanProperty) {
```
Why not using curly brackets here?

recaf-ui/src/main/java/me/coley/recaf/ui/pane/OutlinePane.java
```diff
@@ -31,22 +39,154 @@
...
+		final String icon;
+		final String key;
+
+		MemberType(String icon, String key) {
+			this.icon = icon;
+			this.key = key;
+		}
+
+		@Override
+		public String getTranslationKey() {return key;}
```
NoJokeFNA, line 68:
```suggestion
		public String getTranslationKey() {
			return key;
		}
```

NoJokeFNA (changes requested):

recaf-ui/src/main/java/me/coley/recaf/ui/control/config/ConfigBoolean.java
narumii: XD

Kaspek2480: Rly xD

recaf-core/src/main/java/me/coley/recaf/code/ClassInfo.java
Kaspek2480: Rly xD

recaf-ui/src/main/java/me/coley/recaf/ui/control/code/BracketTracking.java
```diff
@@ -340,12 +340,12 @@ private void addPair(BracketPair pair) {
 		int len = editor.getLength();
 		// Skip if the pair is not in range.
 		if (pair.getStart() >= len - 1 || pair.getEnd() >= len) {
-			logger.trace("Skip add pair {}, not in document bounds", pair);
+//			logger.trace("Skip add pair {}, not in document bounds", pair);
```
Amejonah1200, line 343: Note to myself: revert this

Amejonah1200:
recaf
This the currently implemented new QoL for the outline.
This is a release candidate, what the feature set is concerned. But there are some more things to do: Jetbrains Design Notice for the icons, and also testing, documentation and review.

recaf-core/src/main/java/me/coley/recaf/code/CommonClassInfo.java
```diff
@@ -43,6 +43,18 @@ public interface CommonClassInfo extends ItemInfo {
...
+	 * @return Class's inner classes.
+	 */
+	List<InnerClassInfo> getInnerClasses();
+
+	/**
+	 * Breadcrumbs of the outer class. For example: A, B for the class C which has B as an outer class.
+	 * This List <strong>MUST</strong> be sorted in order of the outest first. The last element is the outer of the class itself.
+	 * @return breadcrumbs of the outer class.
+	 */
+	List<String> getOuterClassBreadcrumbs();
```
Col-E, line 56: Seems to be UI centric data, while this should be strictly a model of the class data.
Breadcrumbds logic can be moved/stored elsewhere.

recaf-core/src/main/java/me/coley/recaf/code/InnerClassInfo.java
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	public String getInnerName() {
+		return innerName;
+	}
+
+	public int getAccess() {
+		return access;
+	}
+}
```
Col-E, line 37: Generally considering how this attribute can be confusing at first, adding some notes on _"this may seem odd but entries like `xyz` can exist when `<condition>`"_ when adding documentation later

recaf-ui/src/main/java/me/coley/recaf/config/ConfigRegistry.java
```diff
@@ -47,6 +49,10 @@ public class ConfigRegistry {
 		registerTypeAdapter(builder, WritableLongValue.class, (w, v) -> w.value(v.get()), r -> new SimpleLongProperty(r.nextLong()));
 		registerTypeAdapter(builder, WritableBooleanValue.class, (w, v) -> w.value(v.get()), r -> new SimpleBooleanProperty(r.nextBoolean()));
 		registerTypeAdapter(builder, WritableStringValue.class, (w, v) -> w.value(v.get()), r -> new SimpleStringProperty(r.nextString()));
+		// TODO: make enum adapters dynamically
```
Col-E, line 52: Are they already not handled? For instance `WorkspaceAction` usage in `DisplayConfig`

Col-E (approved):
Outside of the remaining checkboxes, only a few minor questions

recaf-core/src/main/java/me/coley/recaf/code/CommonClassInfo.java
Amejonah1200: There are 3 ways for getting all outer classes as a breadcrumb:
- getting from inner classes attribute, and then build up the chain (the actual impl)
  the reason why it's only strings not a pair of strings (internal names) and access flags, is because of the root is not there, so it's not possible to get the access flags for it while reading the class.
- splitting by $ and build up the chain (I suppose that's not a good idea because obfuscation can make this difficult?)
- search for all classes in the same package (or whole workspace? I don't know if you can put your inner class/outer class somewhere else) and slowly getting it build up.

So, I don't know if it's a good idea to build the breadcrumbs "externally" from the class read.

recaf-ui/src/main/java/me/coley/recaf/config/ConfigRegistry.java
Amejonah1200: These are enums in a ObjectProperty, so Gson cannot handle them, as it does not know the type of enum it's expecting. (I should probably use the new annotation for that?)

Amejonah1200:
I thought I committed the case sensitivity thing already, seems I went to bed too early and forgot about it. Oops . _.

recaf-ui/src/main/java/me/coley/recaf/config/ConfigRegistry.java
Col-E: Oh gotcha. Good point. Yeah go for it.

recaf-core/src/main/java/me/coley/recaf/code/ClassInfo.java
```diff
@@ -196,31 +198,41 @@ public void visitInnerClass(String name, String outerName, String innerName, int
...
 		// - splitting by $ can go wrong with obfuscation
 		// - searching for outer classes when opening the outline can be slow because if you have a big workspace/package,
 		//   you need to go multiple time through all the entries for reconstruction.
 		String outerClassName = outerClassOf(className, innerClasses);
 		List<String> breadcrumbs = new ArrayList<>();
-		while (outerClassName != null) {
+		int counter = 0; // if some obfuscators think they can trick this by adding many inner classes
+		if (maxOuterDepth > 0) while (outerClassName != null) {
+			if (++counter > maxOuterDepth) {
+				Logging.get(ClassInfo.class)
```
Col-E, line 215: Can extract to constant logger like how it is done in other classes

Col-E (changes requested):

Issue #454, opened 2022-01-24
Assembler's assumption that INVOKEVIRTUAL sets isInterface=false has edge cases


KTibow:
**Describe the bug**

> I was trying to patch a jar with a bug in it, and when I changed the de-assembled bytecode a bit then changed it back to the original, the output didn't work.

(FYI I don't usually do Java, so I basically know nothing, and I'm probably missing something obvious. This probably isn't even your problem, but I don't know what else could be.)
**To Reproduce**

1. Open sample [nocomcrash-1.0.0.zip](https://github.com/Col-E/Recaf/files/7926231/nocomcrash-1.0.0.zip)
2. Navigate to ChatScreenMixin
3. Go to table view, and edit `atomic_interceptChatMessage` with the assembler.
4. Add any line of code, save, remove it, and save.
5. Output bytecode is different than the original, and when you try to use its main functionality (type "=nccrash" into chat on Paper Minecraft 1.18), it crashes.

**Exception**

If applicable, add the exception/stacktrace.
```
[19:39:18] [Render thread/FATAL]: Reported exception thrown!
net.minecraft.class_148: keyPressed event handler
	at net.minecraft.class_437.method_25412(class_437.java:496) ~[client-intermediary.jar:?]
	at net.minecraft.class_309.method_1466(class_309.java:380) ~[client-intermediary.jar:?]
	at net.minecraft.class_309.method_22678(class_309.java:472) ~[client-intermediary.jar:?]
	at net.minecraft.class_1255.execute(class_1255.java:101) ~[client-intermediary.jar:?]
	at net.minecraft.class_309.method_22676(class_309.java:472) ~[client-intermediary.jar:?]
	at org.lwjgl.glfw.GLFWKeyCallbackI.callback(GLFWKeyCallbackI.java:37) ~[lwjgl-glfw-3.2.2.jar:build 10]
	at org.lwjgl.system.JNI.invokeV(Native Method) ~[lwjgl-3.2.2.jar:build 10]
	at org.lwjgl.glfw.GLFW.glfwPollEvents(GLFW.java:3101) ~[lwjgl-glfw-3.2.2.jar:build 10]
	at com.mojang.blaze3d.systems.RenderSystem.flipFrame(RenderSystem.java:195) ~[client-intermediary.jar:?]
	at net.minecraft.class_1041.method_15998(class_1041.java:310) ~[client-intermediary.jar:?]
[...]

PR #591, opened 2022-08-26
Introduce "itf" suffix to invokestatic calls

Earthcomputer:
## What's new

* Introduced an "itf" suffix to invokestatic calls to preserve the fact that it is calling an interface method, which is required by the JVM verifier.

## What's fixed

* #454


Nowilltolife:
That fix is for 2x, 3x uses a new assembler system where the itf problem is solved with a completly custom instruction `invokexinterface`, but now that you remind me i forgot to add `invokestaticinterface`

Nowilltolife:
Forwarded to 3x in #592

Col-E:
A good way to address #454 for 2X without being intrusive to the existing format, nice 👍

PR #596, opened 2022-08-30
Blinking no more in workspace outline for files and classes

Amejonah1200:
## What's fixed
- https://github.com/Col-E/recaf-3x-issues/issues/106

Co-authored by: @xxDark


recaf-core/src/main/java/me/coley/recaf/code/FileInfo.java
```diff
@@ -50,4 +52,17 @@ public byte[] getValue() {
 	public String getExtension() {
 		return extension;
 	}
+
+	private Boolean isText;
```
Col-E, line 56: Prefer to have fields adjacent.

PR #598, opened 2022-09-01
Unreferenced class detection

yapht:
## What's new

* Classes are now aware of their source type (primary, library, internal library, phantom generated)
* Add system for tracking classes that are referenced by code that will execute at runtime, this allows us to show a warning for unused classes when you try to decompile them.


recaf-ui/src/main/java/me/coley/recaf/ui/control/code/java/JavaArea.java
```diff
@@ -243,7 +243,10 @@ public SaveResult save() {
 			if (result.wasSuccess()) {
 				result.getValue().forEach((name, value) -> {
 					logger.info("Updating '{}'", name);
-					resources.getPrimary().getClasses().put(ClassInfo.read(value));
+					ClassSourceType type = ClassSourceType.PRIMARY;
+					if(getCurrentClassInfo() != null)
+						type = getCurrentClassInfo().getSourceType();
```
Col-E, line 248: Saving is disabled for non-primary classes, so this should never happen. Can inline to use constant `PRIMARY` value.

recaf-ui/src/main/java/me/coley/recaf/ui/dialog/SsvmOptimizeDialog.java
```diff
@@ -106,7 +106,7 @@ public SsvmOptimizeDialog(CommonClassInfo owner, MethodInfo info, SsvmIntegratio
 				});
 				byte[] modified = writer.toByteArray();
 				// Replace in workspace
-				WorkspaceAPI.getPrimaryResource().getClasses().put(owner.getName(), ClassInfo.read(modified));
+				WorkspaceAPI.getPrimaryResource().getClasses().put(owner.getName(), ClassInfo.read(modified, owner.getSourceType()));
```
Col-E, line 109: Optimize will only ever operate on primary classes. Can inline and use `PRIMARY`.

recaf-ui/src/main/java/me/coley/recaf/ui/pane/DecompilePane.java
```diff
@@ -123,6 +126,22 @@ private Node createButtonBar() {
 		Label decompilersLabel = new Label("Decompiler: ");
 		box.getChildren().add(decompilersLabel);
 		box.getChildren().add(decompilerCombo);
+
+		if (commonClassInfo != null &&
+				Configs.display().showUnreferencedClassWarning &&
+				commonClassInfo.getSourceType() != ClassSourceType.INTERNAL_LIBRARY && // don't show warning for internal libraries
```
Col-E, line 132: This will allow the warning to show in phantom classes, which is another reason why the phantom enum value can be merged into `INTERNAL_LIBRARY` tbh.

Col-E (changes requested):
Looking good, few minor comments thus far.

Col-E:
Some interesting ideas to revisit on 4.x later, can implement as a `WorkspaceProcessor`. Closing as this targets 3.x and hasn't been updated in 2 years.

PR #601, opened 2022-09-04
Fix local classes and add outer method in breadcrumbs

Amejonah1200:
## What's new

* Outer method in breadcrumbs
* Anonymous and method local classes in class outline

## What's fixed

* https://github.com/Col-E/recaf-3x-issues/issues/110


Col-E:
image

visual

PR #606, opened 2022-09-25
Better suggestions in ASM Editor [WIP]

Amejonah1200:
image

- [x] Sort suggestions based on proximity of the findings: yer > **y**our/proj**e**ct/**r**ight


recaf-ui/src/main/java/me/coley/recaf/ui/control/code/bytecode/AssemblerArea.java
```diff
@@ -295,10 +292,10 @@ private VirtualizedContextMenu<Suggestion> createSuggestionsMenu(int position, G
 					&& e.getInputEvent() instanceof KeyEvent
 					&& ((KeyEvent) e.getInputEvent()).getCode() == KeyCode.TAB) {
 				replaceText(suggestionGroup.start().getStart() + 1, suggestionGroup.start().getEnd() + 1,
-						TextDisplayUtil.escapePath(suggestionText));
+						EscapeUtil.escape(suggestionText));
```
Col-E, line 295: Was renamed to `shortenEscape`.

Should use `TextDisplayUtil` for consistency of using it when possible.

Amejonah1200: What method should I use? As this use case needs to have a path compliant with ASM language.

Nowilltolife: For the text to be compliant with the input transformer all whitespaces need to be formatted which the method `EscapeUtil.escapeSpace` took care of this, but because we take the liberty of saying that we never need to deal with obfuscated input / malformed names.

Col-E: > As this use case needs to have a path compliant with ASM language.

Ah..... yeah. This is correct then.

PR #607, opened 2022-09-25
Add live search to searches [WIP]

Amejonah1200:
recaf
TODO:
- [ ] Use Virtualized Context
- [ ] Implement for other searches
- [ ] Make Result also contain where specifically the match occurred (useful for RegEx)

### DIFF
--- a/recaf-core/src/main/java/me/coley/recaf/search/result/Result.java
+++ b/recaf-core/src/main/java/me/coley/recaf/search/result/Result.java
@@ -1,6 +1,7 @@
 package me.coley.recaf.search.result;
 
 import java.util.Objects;
+import java.util.UUID;
 
 /**
  * The base result contains location information of the matched value.
@@ -35,14 +36,6 @@ public abstract class Result implements Comparable<Result> {
 	}
 
 	@Override
-	public int compareTo(Result o) {
-		if (o == this) {
-			return 0;
-		}
-		return location.compareTo(o.location);
-	}
-
-	@Override
 	public String toString() {
 		return "Result{value=" + getValue() + ", Location=" + location + '}';
 	}
@@ -58,5 +51,19 @@ public abstract class Result implements Comparable<Result> {
 	@Override
 	public int hashCode() {
 		return location.hashCode();
+	}
+
+	private final UUID uuid = UUID.randomUUID();
+
+	@Override
+	public int compareTo(Result o) {
+		if (o == this) {
+			return 0;
+		}
+		int cmp = location.compareTo(o.location);
+		if (cmp == 0) {
+			return uuid.compareTo(o.uuid);
+		}
+		return cmp;
 	}
 }
--- a/recaf-core/src/main/java/me/coley/recaf/search/result/Result.java
+++ b/recaf-core/src/main/java/me/coley/recaf/search/result/Result.java
@@ -10,6 +10,7 @@ import java.util.UUID;
  */
 public abstract class Result implements Comparable<Result> {
 	private final Location location;
+	private final UUID uuid = UUID.randomUUID();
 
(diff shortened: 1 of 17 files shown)


recaf-core/src/main/java/me/coley/recaf/search/result/Result.java
```diff
@@ -59,4 +52,18 @@ public boolean equals(Object o) {
 	public int hashCode() {
 		return location.hashCode();
 	}
+
+	private final UUID uuid = UUID.randomUUID();
```
Col-E, line 56: Format: Field down below like this is weird, should be with the others

recaf-ui/src/main/java/me/coley/recaf/ui/docking/RecafDockingManager.java
```diff
@@ -109,8 +109,8 @@ public static RecafDockingManager getInstance() {
 	}
 
 	@Override
-	protected DockTabFactory decorateFactory(DockingRegion region, DockTabFactory factory) {
-		DockTab tab = factory.get();
+	protected <T extends DockTab> DockTabFactory<T> decorateFactory(DockingRegion region, DockTabFactory<T> factory) {
```
Col-E, line 112: Nit: Above it is `DT`, but here it is just `T`

recaf-ui/src/main/java/me/coley/recaf/ui/pane/SearchPane.java
```diff
@@ -87,9 +99,84 @@ public static SearchPane createTextSearch(String text) {
...
+				(e -> CommonUX.openMember(classInfo, member)) :
+				(e -> CommonUX.openClass(classInfo)));
+			box.setSpacing(6);
+			return box;
+		}
+		return new Label(location.toString());
+	}
+
+	private Instant lastTyped = Instant.EPOCH;
+	private CompletableFuture<?> future = null;
```
Col-E, line 178: Format: Fields not with others at top

recaf-ui/src/main/java/me/coley/recaf/ui/pane/SearchPane.java
```diff
@@ -294,4 +371,81 @@ private void searchDeclaration(String owner, String name, String desc, TextMatch
...
+		}
+
+		@Override
+		protected void updateItem(Result item, boolean empty) {
+			super.updateItem(item, empty);
+			//						updateSelected(false);
+			setGraphic(null);
+			setText(null);
+			if (empty || !(item instanceof TextResult))
+				return;
```
Col-E, line 424: Generally format of `updateItem` used in the project is:

```java
if (empty || item == null) {
	setGraphic(null);
	setText(null);
} else {
	// standard population logic
}
```
This approach uses one less call to `setX` when a cell is recycled to fill something else vs resetting before each usage as the current logic does here.

recaf-ui/src/main/resources/style/seemless-list-view.css
```diff
@@ -0,0 +1,7 @@
+.indexed-cell:even {
+    -fx-background-color: rgb(50, 50, 50);
+}
+
+.indexed-cell:odd {
+    -fx-background-color: rgb(50, 50, 50);
+}
```
Col-E, line 7: Generally unique components all of a single type get put into a single css file _(Note how the `text.css` covers a variety of cases)_.

So this would instead go into `list.css` instead of being its own file.

Col-E (changes requested):

Col-E:
Skipping a few other formatting comments, as I'll just apply the formatter when merging.

Col-E:
Live search has been adapted into the 4.x branch

PR #609, opened 2022-10-10
Fix zoom in PannableImageView

Nowilltolife:
## What's fixed
- Fixed small zoom bug that sometimes occurs

### DIFF
--- a/recaf-ui/src/main/java/me/coley/recaf/ui/control/PannableImageView.java
+++ b/recaf-ui/src/main/java/me/coley/recaf/ui/control/PannableImageView.java
@@ -64,6 +64,7 @@ public class PannableImageView extends BorderPane {
 		});
 		setOnScroll(e -> {
 			double zoomSpeed = 0.05;
+			if(e.getDeltaY() == 0) return; // sometimes happens, but since change is 0, we can ignore it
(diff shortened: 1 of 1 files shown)


recaf-ui/src/main/java/me/coley/recaf/ui/control/PannableImageView.java
```diff
@@ -64,6 +64,7 @@ public PannableImageView(Image image) {
 		});
 		setOnScroll(e -> {
 			double zoomSpeed = 0.05;
+			if(e.getDeltaY() == 0) return; // sometimes happens, but since change is 0, we can ignore it
```
Col-E, line 67: What is the problem here? Should be explained in the comment rather than a generalized statement.

Nowilltolife: Is this better?
```suggestion
			// in some environments the scroll events are prefixed by an event with a delta of 0 before the actual event
			// which causes the below logic to assume it is a zoom out.
			// Since the change is 0 we can just ignore it and return here.
			if(e.getDeltaY() == 0)
				return;
```

Col-E: Yes 👍

PR #611, opened 2022-10-13
Assembler class editing

Nowilltolife:
## What's planned
- Full class assembler editing
- Full analysis of multiple members in the same context

## What's done
- [x] Full class editing skeleton
- [x] Analysis engine for more than one member

## To consider
- Current prs like: #606 need to be adapted if this gets merged
- The context menu logic needs to be tweaked as well

Both of these points tie into the problem of needed a 3rd info when getting AST groups via position: the member owner.

### Site notes
Current draft will cause exception because alot of around things depend on the member being a method

image


recaf-assembler/src/main/java/me/coley/recaf/assemble/transformer/JasmToUnitTransformer.java
```diff
@@ -0,0 +1,384 @@
...
+			}
+
+			@Override
+			public void visitNestMember(NestMemberGroup nestMember) throws AssemblerException {
+				classDefinition.addNestMember(content(nestMember.getMemberName()));
+			}
+
+			@Override
+			public void visitModule(ModuleGroup module) throws AssemblerException {
+
```
Col-E, line 119: Module work pending

Nowilltolife: Added in latest commit

PR #619, opened 2022-11-17
fix the fucking gitpacked jasm

0x3C50:
## What's new

+ fixed the fucking gitpack jasm

## What's fixed

+ fixed gradlew not finding the fucking jasm from gitpack because it's spelled the wrong way

### DIFF
--- a/recaf-assembler/build.gradle
+++ b/recaf-assembler/build.gradle
@@ -14,7 +14,7 @@ dependencies {
     api asm
     api asm_tree
 
-    api "com.github.Nowilltolife:Jasm:$jasmVersion"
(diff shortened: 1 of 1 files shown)


Col-E:
The existing artifact works fine and is resolvable on my end, and on CI's end.

https://jitpack.io/com/github/Nowilltolife/Jasm/1.3.6/Jasm-1.3.6.pom

This will be updated to point to 1.4.1 shortly anyways as 611 is completed.

0x3C50:
It works now, previously only worked with it being spelled like "JASM". This is a true jitpack moment.

PR #621, opened 2022-11-22
3X: CDI refactoring to support more dynamic interactions between modules and plugins

Col-E:
# This is a heavy WIP

This description is not yet complete. I'll elaborate more when I have the time.

## Context

Some context, CDI is an extension of DI systems. First off to understand what DI is, [watch this video](https://www.youtube.com/watch?v=ZZ_qek0hGkM).

What does CDI bring in addition to DI?

 - Lifecycle management
 - Bean discovery

Currently this branch is implemented with Weld. This is probably overkill and if possible swapping to a more minimal implementation will be done. The requirements for a replacement are:

 - Some sort of automatic bean discovery
 - Some sort of lifecycle management for beans associated with a Workspace's lifecycle

We can probably make our own minimal lifecycle system. 

## Why?

1. Dependency injection will be used to allow services in `recaf-core` to expose configurations in such a way to be modified by `recaf-ui`
2. Make way for easy injection of features into plugins via `@Inject`. Need something? We'll maintain a documented list of services. Simply inject it via constructor injection and you got it.

## What is being changed?

Mostly the internals surrounding `Controller` and its `Services` value. Services will be split into two categories:

1. Application scoped _(Not tied to a workspace)_
2. Workspace scoped _(Tied to a workspace, destroyed when workspace is closed)_

This model will refactor components using services deemed _"workspace scoped"_ such that they too are destroyed/removed when the workspace closes. In practice, this will look like all open UI elements pertaining to classes and such in a workspace closing, which already happens for the most-part. It will be be tweaked slightly.


Col-E:
Going to close this as the approach for this will be done in a completely different way, so this branch is being scrapped.

I'll create a new branch with the changes when its ready. It will be much larger in scope.

PR #623, opened 2022-11-22
Fix ow2 handle cast towards HandleInfo

Nowilltolife:
## What's fixed

* Fix ow2 handle cast towards HandleInfo


gaddarkumar7447 (commented):
Now Its correct

PR #625, opened 2022-12-10
JASM update with support for records

Nowilltolife:
## What's new

* Records
* Deprecated attribute
* Attributes moved to `Attributable` interface

## What's fixed

* Some inconsistent attribute parsing


recaf-assembler/src/main/java/me/coley/recaf/assemble/ast/arch/Attributable.java
```diff
@@ -0,0 +1,19 @@
+package me.coley.recaf.assemble.ast.arch;
+
+import me.coley.recaf.assemble.ast.meta.Signature;
+
+import java.util.List;
+
+public interface Attributable {
```
Col-E, line 7: I do not understand the relation between the class name, and its capabilities.

recaf-assembler/src/main/java/me/coley/recaf/assemble/ast/arch/AbstractDefinition.java
```diff
@@ -1,65 +1,44 @@
...
-import java.util.List;
 
 /**
  * Common base implementation for definition types.
  *
  * @author Matt Coley
  * @author Justus Garbe
  */
-public abstract class AbstractDefinition extends BaseElement implements Definition {
-	private final List<Annotation> annotations = new ArrayList<>();
```
Col-E, line 17: Auto-formatting replaced `\t` with `    `

recaf-assembler/src/main/java/me/coley/recaf/assemble/ast/arch/ClassDefinition.java
```diff
@@ -194,6 +204,13 @@ public Module getModule() {
 		return module;
 	}
 
+	/**
+	 * @return Record of the class.
+	 */
+	public Record getRecord() {
```
Col-E, line 210: Would be more grammatically correct to say _"Record attributes of the class"_. When this attribute exists, the class _is_ a record. So saying `classThatIsRecord.getRecord()` is odd semantics IMO.

recaf-assembler/src/main/java/me/coley/recaf/assemble/transformer/AnnotationHelper.java
```diff
@@ -30,7 +31,7 @@ public class AnnotationHelper {
 	 * @param annotations
 	 * 		The annotations to add.
 	 */
-	public static void visitAnnos(AbstractDefinition code, boolean visible, List<AnnotationNode> annotations) {
+	public static void visitAnnos(AbstractAttributable code, boolean visible, List<AnnotationNode> annotations) {
```
Col-E, line 34: If the purpose of `Attributable` is to say _"this type has annotations"_ then it should be `Annotatable` or something along those lines.

Col-E:
👍 for new language features supported

Got some design things I'd like to see changed though, as commented above, and auto-format things.

recaf-assembler/src/main/java/me/coley/recaf/assemble/transformer/AnnotationHelper.java
Nowilltolife: The reason for `Attributable`s existance is that there exists a overlap for attributes with `declaration` and `record component` infos according to the [4.7-C Table](https://docs.oracle.com/javase/specs/jvms/se19/html/jvms-4.html#jvms-4.7.20) in the jvm. Now because the signature and annotation fields and spec are trapped within the `Definition` interface i needed to extract it to a `Attributable` (name i came up with at first (dont know if correct)) interface which then interfaces `AbstractAttributable` which is extended by `AbstractDefinition` AND `RecordComponent` and because the method only supported reading annotations into a `AbstractDefinition`  i needed to make it adapt to the attributable and i dont name it `Annotable` because `Attributable` contains more than one attribute in it. While i do acknowledge creating a interface spec for 2 fields can be considered excessive but creating 2 extra interfaces for them is even more

recaf-assembler/src/main/java/me/coley/recaf/assemble/ast/arch/AttributeContainer.java
```diff
@@ -18,8 +18,7 @@ public List<Annotation> getAnnotations() {
 	}
 
 	public void setAnnotations(List<Annotation> annotations) {
-		this.annotations.clear();
-		this.annotations.addAll(annotations);
+		this.annotations = annotations;
```
Col-E, line 21: Why?

Amejonah1200: You need to know the consequences, you cannot "listen" to annotations anymore. Like taking getAnnotations and having a reference only to that. Make sure that this is intended behavior.

PR #645, opened 2023-05-03
Fix exception caused by mismatching types when setting default decompiler

leonardvdj:
## What's fixed

* Fix IllegalArgumentException by setting decompiler field to decompiler name.


Col-E:
👍

PR #664, opened 2023-08-23
Fix toolchain resolving

virustotalop:
## What's new

* N/A

## What's fixed

* Automatic toolchain resolving

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,5 +1,5 @@
 plugins {
-    id 'org.gradle.toolchains.foojay-resolver-convention' version '0.4.0'
+    id 'org.gradle.toolchains.foojay-resolver-convention' version '0.7.0'
 }
 
 rootProject.name = 'recaf'
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,3 +1,7 @@
+plugins {
+    id 'org.gradle.toolchains.foojay-resolver-convention' version '0.4.0'
+}
+
 rootProject.name = 'recaf'
 include 'recaf-utils',
         'recaf-core',


Col-E:
The dev3 branch resolves fine already. Can you explain what the purpose of this plugin is? And why are you using an outdated version when the latest is 0.7.0?

virustotalop:
Good catch with the version, I just pulled the plugin from the [gradle docs](https://docs.gradle.org/8.2.1/userguide/toolchains.html#sec:auto_detection) without checking the latest version. Without the toolchain resolver plugin the Java 8 toolchain doesn't resolve for the launcher module. I'm not sure if that can be updated to 11 and if the issue will be resolved but it is fixed with the toolchain plugin. Without the toolchain plugin the follow error occurs.
```
Build file 'C:\Users\vtotal\Documents\GitHub\Recaf\recaf-launcher\build.gradle' line: 44

* What went wrong:
A problem occurred evaluating project ':recaf-launcher'.
> Failed to calculate the value of task ':recaf-launcher:compileJava' property 'javaCompiler'.
   > No matching toolchains found for requested specification: {languageVersion=8, vendor=any, implementation=vendor-specific} for WINDOWS on x86_64.
      > No locally installed toolchains match and toolchain download repositories have not been configured.

* Try:
> Learn more about toolchain auto-detection at https://docs.gradle.org/8.2.1/userguide/toolchains.html#sec:auto_detection.
> Learn more about toolchain repositories at https://docs.gradle.org/8.2.1/userguide/toolchains.html#sub:download_repositories.
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.
```

Steps to reproduce
1. Navigate to your `.gradle` directory
2. Delete all the cached jdks in the `jdks` directory
3. You should get the error above if you have Java 17 installed and no other jdks installed on the machine

Col-E:
> Without the toolchain resolver plugin the Java 8 jdk doesn't download to be used by the launcher module.

Ah gotcha.

The plan for the launcher is to give that to users who otherwise would run into issues they're not technical enough to resolve themselves. So while 3X requires Java 11, the launcher on Java 8 could be used to warn you that you still need to update, without itself failing.

PR #665, opened 2023-08-24
[Feature] Added New Input Boxes for Config Options and Wrote Swedish Translation

pvpb0t:
#### Added Input Boxes for Config Options
- Introduced a number input box for Config Options that uses type Integer.
- Implemented a String input box for Config Options that uses type String.
bild

#### Translated recaf to Swedish
- Wrote a Swedish translation of Recafs dialogoues

## What's fixed

- No more "Unsupporrted java.lang.Integer/java.lang.String

### DIFF
--- a/recaf-ui/src/main/java/software/coley/recaf/services/config/factories/NumberComponentFactory.java
+++ b/recaf-ui/src/main/java/software/coley/recaf/services/config/factories/NumberComponentFactory.java
@@ -1,0 +1,37 @@
+package software.coley.recaf.services.config.factories;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+import javafx.scene.Node;
+import javafx.scene.control.TextField;
+import software.coley.observables.Observable;
+import software.coley.recaf.config.ConfigContainer;
+import software.coley.recaf.config.ConfigValue;
+import software.coley.recaf.services.config.TypedConfigComponentFactory;
+@ApplicationScoped
+public class NumberComponentFactory extends TypedConfigComponentFactory<Integer> {
+
+    @Inject
+    protected NumberComponentFactory() {
+        super(false, Integer.class);
+    }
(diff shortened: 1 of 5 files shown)


recaf-ui/src/main/java/software/coley/recaf/services/config/factories/NumberComponentFactory.java
```diff
@@ -0,0 +1,37 @@
...
+        String translationKey = container.getScopedId(value);
+        TextField textField = new TextField();
+        textField.setText(Integer.toString(observable.getValue()));
+        textField.textProperty().addListener((observableValue, oldValue, newValue) -> {
+            try {
+                int newValueAsInt = Integer.parseInt(newValue);
+                observable.setValue(newValueAsInt);
+            } catch (NumberFormatException e) {
+                textField.setText(oldValue);
+            }
```
Col-E, line 31: I'll play with the UX later to see if this feels right when in use. May have some odd cases like starting to write a negative number, backspacing to an empty string, etc

recaf-ui/src/main/java/software/coley/recaf/services/config/factories/NumberComponentFactory.java
```diff
@@ -0,0 +1,37 @@
...
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+import javafx.scene.Node;
+import javafx.scene.control.TextField;
+import software.coley.observables.Observable;
+import software.coley.recaf.config.ConfigContainer;
+import software.coley.recaf.config.ConfigValue;
+import software.coley.recaf.services.config.TypedConfigComponentFactory;
+@ApplicationScoped
+public class NumberComponentFactory extends TypedConfigComponentFactory<Integer> {
```
Col-E, line 12: Name should be `IntegerComponentFactory`

PR #718, opened 2023-09-09
4x: Starting to flesh out SSVM integration

Col-E:
Currently waiting on feedback on https://github.com/xxDark/SSVM/pull/27 before moving forward


Col-E:
Did some rebasing on on 4x, opening a new PR with changes off of that changed work #721

Issue #584, opened 2022-08-05
JavaParser issues in '< >'


CXVUSER:
parser fails in:
private WeakIdentityHashMap<IUnit, Integer> interpreterIds; 
![image](https://user-images.githubusercontent.com/61128558/183060934-73420fb0-8e5b-4e7b-a2ed-fe87b76b0372.png)
![image](https://user-images.githubusercontent.com/61128558/183060994-329e3979-2fed-4088-8f67-751a5b11bd34.png)

parser no fails in:
private WeakIdentityHashMap<IUnit, Integer interpreterIds; 

???

JDK 11.0.6
recaf-3.0.0-SNAPSHOT-[3d01eec](https://github.com/Col-E/Recaf/commit/3d01eec563f0532ba9bcead12ba6ca8c5d60925c)

xxDark:
We can't do anything about that the moment (correct me if I'm wrong), but it is recommended that in cases like that you use assembler mode, which the warning suggests.

Col-E:
I'd like to have an example class that reproduces this consistently in order to investigate the cause.

CXVUSER:
`package me.coley.jremapper.ui;

import me.coley.jremapper.util.Logging;
import java.util.Arrays;
import java.util.TreeMap;
import javafx.scene.input.TransferMode;
import me.coley.event.Event;
import me.coley.jremapper.event.ClassOpenEvent;
import javafx.scene.input.Dragboard;
import java.io.File;
import javafx.scene.image.ImageView;
import me.coley.jremapper.util.Icons;
import javafx.scene.control.TreeCell;
import me.coley.jremapper.mapping.CMap;
import me.coley.jremapper.event.MappingChangeEvent;
import java.util.Iterator;
import java.util.zip.ZipEntry;
import java.util.jar.JarEntry;
import java.io.OutputStream;
import java.util.jar.JarOutputStream;
import java.io.FileOutputStream;
import java.util.Map;
import java.util.HashMap;
import me.coley.jremapper.event.SaveJarEvent;
import java.nio.file.Path;
import com.eclipsesource.json.JsonValue;
import java.nio.file.Files;
import java.nio.file.StandardOpenOption;
import java.nio.file.OpenOption;
import java.nio.charset.StandardCharsets;
import com.eclipsesource.json.WriterConfig;
import me.coley.jremapper.event.SaveMapEvent;
import me.coley.jremapper.mapping.Mappings;
[...]

PR #746, opened 2023-11-30
fix: parser removing more than generics

Jydett:
Hello, firstly, thank you for this great tool !

I encounter an issue the other day similar to [[this one](https://github.com/Col-E/Recaf/issues/584)](https://github.com/Col-E/Recaf/issues/584) where the comparison in this pseudo-class would be matched as a generic declaration and removed before being parsed as java code, thus resulting in java parsing errors.

*decompiled code*
```
class Clazz {
   public boolean test(int x) {
      int a = 0, b = 0;
      if (a < x && x > b) {
          return false;
      }
      return true;
   }
}
```

*after generic type erasing*
```
class Clazz {
   public boolean test(int x) {
      int a = 0, b = 0;
      if (a            b) {
          return false;
      }
      return true;
   }
}
```

I tried to fix the regular expression, but it would become massive, hard to understand and inefficient, so I rewrote the function as a simple syntax parser and added some tests.

### DIFF
--- a/recaf-core/src/main/java/me/coley/recaf/parse/JavaParserHelper.java
+++ b/recaf-core/src/main/java/me/coley/recaf/parse/JavaParserHelper.java
@@ -1,21 +1,22 @@
 package me.coley.recaf.parse;
 
 import com.github.javaparser.*;
+import com.github.javaparser.ParserConfiguration.LanguageLevel;
 import com.github.javaparser.ast.CompilationUnit;
 import com.github.javaparser.ast.Node;
 import com.github.javaparser.ast.body.*;
 import com.github.javaparser.ast.expr.Expression;
 import com.github.javaparser.ast.expr.ObjectCreationExpr;
 import com.github.javaparser.ast.expr.SimpleName;
-import jregex.Matcher;
+import java.util.Arrays;
 import me.coley.recaf.Controller;
 import me.coley.recaf.code.FieldInfo;
 import me.coley.recaf.code.ItemInfo;
 import me.coley.recaf.code.LiteralExpressionInfo;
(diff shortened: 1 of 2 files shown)


Jydett:
I'll switch this PR to draft as I discovered some a new edge case that will be hard to fix: 
```
    public static void main(String[] args) {
        int a = 0, b = 0, c = 0, d = 0;
        boolean x = a < b & c > d;
    }
```

meiMingle:
@Jydett Hi, do you think this PR can solve the edge case you mentioned? #748

Jydett:
@meiMingle Hi, i don't think so. If you want to check you can copy [the test class](https://github.com/Jydett/Recaf/blob/fix/parser/recaf-ui/src/test/java/me/coley/recaf/ui/control/code/JavaParserTest.java) to your PR and run it to validate its behaviour

meiMingle:
@Jydett I just used my newly created regular expression to run your test class and the only thing that failed was the following method, `<T extends EntityLivingBase<X> | Foo>` are you sure this syntax is correct? What environment (JDK version/language level) does it come from?
```java
@Test
void removeGenericMethodDeclaration() {
String code = "<T extends EntityLivingBase<X> | Foo> foo() {}";
var res = filterGeneric(code);
assertEquals(" foo() {}", res);
}
```assertEquals(" foo() {}", res);
}
```

Jydett:
@meiMingle You're right the correct syntax is ```<T extends EntityLivingBase<X> & Foo>```

Col-E:
Closing out as we're focusing on 4.x now, and its not using JavaParser as a backend.

PR #756, opened 2024-01-01
fix: infinite loading on launch

demkom58:
## What's fixed

- infinite loading on launch


Col-E:
Woo! Windows & 3rd party ps problems 🥳 

As per your note on the linked issue on newer versions, I'll ensure we skip using the library on windows so this doesn't occur.

PR #761, opened 2024-01-18
perf: 2x speed up reading contents from zip

1fxe:
This specific optimistaions is targeted towards JVM Class resources, in is isAsmCompliantClass we would read the class and check for custom attributes, then we would read the class again to build the class info in this pr I merged the functionality of isASMCompliantClass into ClassBuilderAppender

## What's new

Before ~ 5746ms on a 100mb jar
image

After ~ 2823ms
image


recaf-core/src/main/java/software/coley/recaf/workspace/io/BasicInfoImporter.java
```diff
@@ -44,26 +43,26 @@ public Info readInfo(@Nonnull String name, @Nonnull ByteSource source) throws IO
 		// Check for Java classes
 		if (matchesClass(data)) {
 			try {
-				// Patch if not compatible with ASM
-				if (!isAsmCompliantClass(data)) {
+				try {
+					return new JvmClassInfoBuilder(data)
+						.skipASMValidation(config.doSkipAsmValidation())
+						.build();
+				} catch (IllegalStateException unknownAttribute) {
```
1fxe, line 50: This try ctach stuff is kind of cringe I think ill just change to boolean check

1fxe (commented):
Small thing to change

Col-E:
I oughta look into keeping codecov but not having it "fail" the build...

1fxe:
Found a bug with reading module-info with these changes on no

1fxe:
Lemme do some more testing 🤔

recaf-core/src/main/java/software/coley/recaf/workspace/io/BasicInfoImporter.java
```diff
@@ -43,17 +41,21 @@ public Info readInfo(@Nonnull String name, @Nonnull ByteSource source) throws IO
 		// Check for Java classes
 		if (matchesClass(data)) {
 			try {
+				if (config.doSkipAsmValidation()) {
```
1fxe, line 44: Since its not just for checking custom attributes

1fxe (commented):
Ok looks good to me

PR #765, opened 2024-02-17
Fix jitpack

crazycat256:
## What's new

Nothing

## What's fixed

Jitpack
By default, jitpack uses java 8, this PR adds a `jitpack.yml` that tells it to use java 17

### DIFF
--- a/jitpack.yml
+++ b/jitpack.yml
@@ -1,0 +1,5 @@
+before_install:
+  - source "$HOME/.sdkman/bin/sdkman-init.sh"
+  - sdk update
+  - sdk install java 17.0.1-zulu
+  - sdk use java 17.0.1-zulu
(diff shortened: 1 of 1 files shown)


0x3C50:
sdkman init and update can be skipped, jitpack does those two steps automatically. Take a look at the [renderer jitpack.yml](https://github.com/0x3C50/Renderer/blob/master/jitpack.yml)

crazycat256:
I've removed these 2 steps

Col-E:
@0x3C50 It looks this is incorrect, as removing this results in a failed build on JitPack.

https://jitpack.io/com/github/Col-E/Recaf/dev4-45a626cce9-1/build.log

0x3C50:
> @0x3C50 It looks this is incorrect, as removing this results in a failed build on JitPack.
> 
> https://jitpack.io/com/github/Col-E/Recaf/dev4-45a626cce9-1/build.log

the build works on renderer. try openjdk?

Col-E:
Ah I see the discrepancy, in renderer the `jdk:` block is configured rather than relying on `sdk update`

PR #768, opened 2024-03-17
Feature/dependency catalogs

Amejonah1200:
## What's new

* Update Gradle Wrapper to 8.6
* Migrate from dependencies.gradle to gradle/libs.versions.toml to reduce the need to recompile if versions are changed


Amejonah1200:
image
🤨

PR #775, opened 2024-04-11
JDK 22 migraiton

xxDark:
## What's new

Migrate to JDK 22. Blocked on https://github.com/Col-E/LL-Java-Zip/pull/28

Pros:
- Newest API
- Can upgrade JavaFX

Cons:
- Not LTS
- Will probably have to do something about these nasty warns in `ReflectUtil` at some point..
- Some people would cry


Col-E:
gradle complaining

gradle rel notes

huh?

Issue #257, opened 2020-08-29
[Suggestion] Warn when changing the keybind for 'Undo changes' to Ctrl-Z


andylizi:
I was fiddling around with the settings when I saw the default hotkey for undo is `Ctrl-U`, and I thought to myself "Oh, and why would any hotkey that undo stuff not be `Ctrl-Z`?" and properly decided to *correct* this oversight. Later when I was writing lines and lines of instruction in the assembler, I absentmindedly hit `Ctrl-Z` *AND ALL THE CHANGES I DID WERE GONE IN AN INSTANT -* 

Please show a warning when the keybind is changed to `Ctrl-Z`. Or maybe clarify just what **exactly** does *"most recent changes"* in the phrase "undo the most recent changes in the current editor" mean. Or maybe add a redo action.

Col-E:
😂 yeah, thats exactly why that is bound that way

PR #783, opened 2024-04-29
#257 Warn before setting undo keybind to ctrl-z

lbroglio:
## Description
After looking at the issue #257 I added a check when changing keybinds that the keybind being set is undo and its being changed to ctrl-z a prompt will appear asking the user to confirm the change and explaining this isn't the standard text edit undo.

## Commit

feat: A confirmation prompt now appears when changing to undo keybind to ctrl-z

## What's new

* If the undo keybind is set to ctrl-z the user will be asked to confirm the decision by a pop up


xxDark:
Hello. 2.x is no longer supported. Please target dev4 branch if this change is relevant to 4.x code base.

Issue #580, opened 2022-08-01
Shortcuts


RandomVersion:
## Shortcuts

For example pressing f2 with open the rename dialog and so on.

Amejonah1200:
Can you please create a list of all shortcuts you wish?

xcfrg:
All the ones from IDA :trollface: 
`n` should be rename, not f2

RandomVersion:
make it configurable!

RandomVersion:
maybe add like presets
for example the shortcuts from ida as @xcfrg said and others

Amejonah1200:
Please create a list of all shortcuts you like to have implemented. So it's possible to plan out.

PR #784, opened 2024-04-29
#580 Add new alternate rename hotkey so F2 can be mapped by default

lbroglio:
## Description
After reviewing issue #580 I saw that the current default rename hotkey (ctrl-r) can't be kept while mapping a function key (the issue suggests F2) to rename. I added a second hotkey for rename which allows for a function key (by default F2) to be mapped to rename.

## Commit
feat: Setup F2 as an alternate hotkey for rename

## What's new

* A second rename hotkey has been added which is set to F2 by default


xxDark:
Hello. 2.x is no longer supported. Please target dev4 branch if this change is relevant to 4.x code base.

Issue #642, opened 2023-04-04
Translate to Spanish


airvzxf:
## Translate to Spanish

Add a translation to Spanish, adding a new file in `src/main/resources/translations/es.json`.

Col-E:
Fair warning, a lot of translations in the `master` branch will become irrelevant once the rewrite work is merged back in.

airvzxf:
Thanks, I started to translate it but with this information, I am stopping the work. At what moment do you think is good to start the translation? Or would it be possible to start to translate in the `dev3` branch?

Col-E:
A lot of the strings in the `dev3` branch will be retained so that would be a safer starting point.

PR #785, opened 2024-04-30
Fixed issue #642 by adding a json file for Spanish translation

SpencerToddT:
## What's new

Added file sp.json with spanish translations.

## What's fixed

Only additions, no fixes.


Col-E:
Unfortunately this is targeting 2.X which we are no longer updating :(

PR #812, opened 2024-06-25
feat: detect lookup switch labels in search results

Jydett:
## What's new

Allow the search to match switch labels, and allow the result to be displayed in the result panel as 'lookupswitch'.

In my usecase, I have a huge codebase with a lot of 'switch' but I can't search for their  labels in Recaf 4.X (they are number):
image

Results:
image
Without the edit in BlwUtil, the instruction is passed to ``execute(Instruction instruction)` and thus nothing is displayed.
image

### DIFF
--- a/recaf-core/src/main/java/software/coley/recaf/util/BlwUtil.java
+++ b/recaf-core/src/main/java/software/coley/recaf/util/BlwUtil.java
@@ -90,6 +90,8 @@ public class BlwUtil {
 				printer.execute(constWrapped);
 			else if (wrapped instanceof PrimitiveConversionInstruction convWrapped)
 				printer.execute(convWrapped);
+			else if (wrapped instanceof SimpleInstruction simpleWrapped)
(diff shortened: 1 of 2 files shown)


recaf-core/src/main/java/software/coley/recaf/util/BlwUtil.java
```diff
@@ -90,6 +90,8 @@ else if (wrapped instanceof AllocateInstruction allocateWrapped)
 				printer.execute(constWrapped);
 			else if (wrapped instanceof PrimitiveConversionInstruction convWrapped)
 				printer.execute(convWrapped);
+			else if (wrapped instanceof SimpleInstruction)
+				printer.execute((SimpleInstruction) wrapped);
```
Col-E, line 94: For consistency use the inline `instanceof` pattern matching.

Col-E:
👍

PR #827, opened 2024-07-16
Text overlay support (such as error indicator squiggly lines)

Col-E:
Creates a more standard API for paragraph factories to render on top of the `Editor`

image


recaf-ui/src/main/java/software/coley/recaf/ui/control/richtext/problem/Problem.java
```diff
@@ -8,38 +8,11 @@
  *
  * @author Matt Coley
  */
-public class Problem implements Comparable<Problem> {
-	private final int line;
-	private final int column;
-	private final ProblemLevel level;
-	private final ProblemPhase phase;
-	private final String message;
+public record Problem(int line, int column, int length, ProblemLevel level, ProblemPhase phase,
```
Col-E, line 11: There are no longer javadocs for the parameters of the constructor. The old constructor docs for `@param` use need to be copied to the class docs now that this is a record.

PR #829, opened 2024-07-21
feat: Added simple control flow graph view 

Krystian20857:
## What's new

Added simple control flow graph view.

image


Col-E:
For those looking for control flow graphing, I do intend to eventually support this, but not by bundling any Swing-based libraries and usage through compatibility layers. For reference, in [AreaAnalysisPane](recaf-ui/src/main/java/software/coley/recaf/ui/pane/analysis/AreaAnalysisPane.java) usage of `SpatialCanvas`  shows how to build a graph, specify a layout engine + renderers and display the result. Something along those lines will be what I'm looking at in the future.

Issue #866, opened 2024-10-18
Making it possible for vineflower and fernflower to coexist


meiMingle:
## Feature_Name

Use the Relocating Packagesg function of the com.github.johnrengelman.shadow plugin to process the decompiler class library. This can theoretically allow vineflower and fernflower and other similar decompilers to coexist in Recaf. This has the following benefits: 
① Recaf users can get more compiler results as references 
② Vineflower and fernflower and other similar decompilers developers can compare the differences between their decompilation results, promoting the update and iteration of decompilers

Col-E:
The issue is compiling against both. They both share the same names, so even if we add the dependency for both on the classpath, shadow will just map both of them to whatever is being shaded since it cannot tell the difference.

meiMingle:
This may be of reference value https://github.com/Earthcomputer/class-file-indexer/blob/main/build.gradle.kts

meiMingle:
I think it makes sense to do this, and I will work hard on it. At present, I have completed some work and realized the coexistence of fernflower, vineflower, and forgeflower libraries. The next step is to implement 'AbstractJvmDecompiler` and `BaseDecompilerConfig` for fernflower and forgeflower. Wait for my good news! !

meiMingle:
Currently, rough support for fernflower and forgeflower has been completed. The next step is to sort out the relevant configuration items and complete the configuration interface. Wait for my good news!!

PR #871, opened 2024-10-29
Make  _flower series of decompilers coexist

meiMingle:
As the title says, this is to implement #866. This PR refers to the method in [class-file-indexer](https://github.com/Earthcomputer/class-file-indexer) to relocate the package path of the _flower series decompiler to achieve the purpose of coexistence without interfering with each other.


gradle/libs.versions.toml
```diff
@@ -128,6 +132,12 @@ treemapfx = { module = "software.coley:treemap-fx", version.ref = "treemapfx" }
 
 vineflower = { module = "org.vineflower:vineflower", version.ref = "vineflower" }
 
+fernflower = { module = "com.jetbrains.intellij.java:java-decompiler-engine", version.ref = "fernflower" }
+
+forgeflower = { module = "net.minecraftforge:forgeflower", version.ref = "forgeflower" }
+
+annotations = { module = "org.jetbrains:annotations", version.ref = "annotations" }
```
Amejonah1200, line 139: Why were JetBrains annotations added?

recaf-core/build.gradle
```diff
@@ -38,7 +38,8 @@ dependencies {
     api(libs.openrewrite)
     api(libs.regex)
     api(libs.bundles.jasm)
-    api(libs.vineflower)
+//    api(libs.vineflower)
+    api( project(':recaf-relocation'))
```
Col-E, line 42: I know that you put in a fair amount of effort to get these decompilers to all work with one another, but I am rather against having such things like this relocation module in the main repository.

Additionally, the amount of differences between these decompilers is rather small since they are all forked off one-another. 

IMO these alternative flowers should be offered as plugins which are independently shading their respective flower code into the plugin jars. Decompiler implementations and config containers can be registed on the fly so there's no concern with lessened capability by having them offered separately.

Plus, given some of our user interactions I have a strong feeling that having nearly the same decompiler three times would be confusing to them.

gradle/libs.versions.toml
meiMingle: This is a dependency of fernflower and forgeflower, and may not be a necessary dependency. I just tested it and it has no effect if I remove it.

meiMingle: I will do more testing and decide whether to remove it.

recaf-core/build.gradle
meiMingle: > Additionally, the amount of differences between these decompilers is rather small since they are all forked off one-another.

Although I know that they are forked from each other, I cannot agree that their differences are small. In fact, after years of evolution, they are now very different in terms of interface form, number of controllable parameters, and compilation results.
In terms of the number of parameters, vineflower: 64, fernflower: 45, forgeflower: 47.
The following three figures are the decompilation results of the three flower decompilers for `software.coley.instrument.io.ByteBufferSanitizer` in `%APPDATA%/Recaf/agent/agent-1.4.2.jar` in the Recaf running directory. The source code of the corresponding class is located in [ByteBufferSanitizer](https://github.com/Col-E/InstrumentationServer/blob/master/src/main/java/software/coley/instrument/io/ByteBufferSanitizer.java).  Among them, only fernflower obtained the equivalent result to the source code.
Vineflower
Fernflower
Forgeflower

When decompiling some internal enumeration classes, vineflower reports an error, while fernflower and forgeflower can get the decompiled results. Below are the decompiled results of the three flower class decompilers for `jadx.api.JadxArgs$RenameEnum` in [jadx-core-1.5.0.jar](https://repo1.maven.org/maven2/io/github/skylot/jadx-core/1.5.0/jadx-core-1.5.0.jar)

Vineflower-1
Fernflower-1
Forgeflower-1





> IMO these alternative flowers should be offered as plugins which are independently shading their respective flower code into the plugin jars. Decompiler implementations and config containers can be registed on the fly so there's no concern with lessened capability by having them offered separately.

Maybe this is one of the possible solutions, but I don't know how to create a plug-in for Recaf yet. Is there any sample for reference?

> Plus, given some of our user interactions I have a strong feeling that having nearly the same decompiler three times would be confusing to them.

For ordinary users, they mostly use the default CFR decompiler and use the default configuration. For professional users, they should know clearly which compiler they want to use and what configuration parameters they want to use. Maybe I need to further optimize the configuration name to make it more clearly match the parameter name in the corresponding decompiler official documentation.

gradle/libs.versions.toml
meiMingle: I have tested many times, and the dependence has no effect. Now he has been removed.

meiMingle:
Following @Col-E 's suggestion, this PR will be provided as a separate plugin [fernflower-recaf-plugin](https://github.com/meiMingle/fernflower-recaf-plugin) Currently this plugin is initially available, but there are still some issues that need to be resolved.

Issue #858, opened 2024-09-22
Auto-scroll with middle mouse button


b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0:
## Subject: Feature request: Auto-scroll with middle mouse button

Hey there!

I’d love to see an auto-scroll feature added to Recaf. You know, the one where you press the middle mouse button, and then you can scroll by just moving the mouse up or down? It would make it way easier to go through long lists or sections in the program without having to keep spinning the scroll wheel.

I think this would be a great quality of life addition and super helpful for users like me.

Thanks for considering it, and thanks for all the awesome work you’ve done on Recaf!

PR #890, opened 2024-12-03
Auto-scroll with middle mouse button

funcid:
## What's new

Closes https://github.com/Col-E/Recaf/issues/858

## What's fixed

None


Col-E:
Moved the scroll behavior to the wrapping scroll-pane so other controls can benefit from the new behavior:

https://github.com/user-attachments/assets/390715d2-0678-4533-bd17-431216593c97

Thank you for the contribution 👍

PR #903, opened 2025-01-08
Fix EnigmaMappings parsing

Janmm14:
## What's new

- support `-` as no mapping
- support skipping over `ACC:xxx` access modifier
- 3x to 5x speed improvement in parsing (tested using https://discord.com/channels/443258489146572810/1166247896384028804/1325906402304327690)

## What's fixed

- inner classes are now prefixed with outer class(es)

The parser now only supports tab indents (as per enigma mappings specification) , no longer other characters trim is removing (ascii code points < 19 no longer regarded as tab.


Col-E:
Had some local changes which confused github, but this is merged now 👍

PR #905, opened 2025-01-09
Enigma mappings improve 2

Janmm14:
Rewrote the parser.

- Still using indices for fast string parsing
- Fixed inner classes again
- Should be able to create better error messages
- Improved tolerance for bogus spaces
- Basic indent validation
- No state machine, state is handled by code flow

Please give feedback on complexity and maintainability.

~~It seems to a little faster than the previous one.~~ Speed about the same as the previous new impl.
Due to lambda usage `MemberMappingsConsumer` for code clarity, first shot speed is a little slower.


Col-E:
> Please give feedback on complexity and maintainability.

Compared to the prior state-machine based system, this is much easier. I was able to read through the utility methods and easily fill in doc comments.

PR #941, opened 2025-06-10
Replace docking framework

Col-E:
# UI Comparison

Left is with Bento. Right is the existing solution with TiwulFX.

image

# New features

Clicking on tabs that occupy the full width/height of a split will collapse them. This lets you free up space quickly while keeping those tabs one click away.

image

Where things will get split should be more intuitive now. Here is another comparison when dragging the `Bootstrap` class.

image


recaf-ui/src/main/java/software/coley/recaf/ui/docking/DockingLayoutManager.java
```diff
@@ -0,0 +1,221 @@
...
+									ContextMenu menu = new ContextMenu();
+									addSideItems(menu, space);
+									return menu;
+								})
+				))
+		);
+	}
+
+	private static void addSideItems(@Nonnull ContextMenu menu, @Nonnull TabbedDockSpace tabbed) {
+		// TODO: Move this to common location, and make other things use it
```
Col-E, line 134: I've not yet figured out where would be best to put this. Mainly since I dont know how many places I want it to be used in. I don't want every tabbed space to have the side being adjustable, so I can't make it the default...

recaf-ui/src/main/java/software/coley/recaf/ui/docking/DockingManager.java
```diff
@@ -24,230 +31,100 @@
...
-		// Needed in case a window containing the region gets closed.
-		for (DockingTab tab : new ArrayList<>(region.getDockTabs()))
-			tab.close();
-
-		// Tell the region it is closed, removing its reference to this docking manager.
-		region.onClose();
-
-		// Closure allowed.
-		return true;
+	public boolean replace(@Nonnull String identifier, @Nonnull SpaceSupplier supplier) {
```
Col-E, line 55: Need to add back Javadocs

recaf-ui/src/main/java/software/coley/recaf/ui/pane/editing/AbstractContentPane.java
```diff
@@ -109,20 +165,63 @@ protected void refreshDisplay() {
...
-		// Lazily create/add side-tabs to UI.
-		if (sideTabs == null) {
-			sideTabs = new SideTabs(Orientation.VERTICAL);
-			children.add(sideTabs);
-			setRight(sideTabs);
+	public void addSideTab(@Nonnull ObservableValue<String> binding, @Nonnull Ikon icon, @Nullable Node content) {
+		addSideTab(binding, d -> new FontIconView(icon), content);
+	}
+
+	public void addSideTab(@Nonnull ObservableValue<String> binding, @Nonnull DockableIconFactory iconFactory, @Nullable Node content) {
```
Col-E, line 172: Javadocs here and other public methods that are missing docs.

recaf-ui/src/main/resources/style/recaf.css
```diff
@@ -432,7 +432,6 @@ Text {
   -fx-graphic-text-gap: 6px;
   -fx-text-fill: -color-button-fg;
   -fx-alignment: CENTER;
-  -fx-effect: dropshadow(gaussian, -color-button-shadow, 3px, -2, 0, 1);
```
Col-E, line 435: Removed because it would cause the buttons used in Bento's header view to infinitely expand

PR #942, opened 2025-06-23
Add an option for analysis when opening the workspace

luiox:
## What's new

* Move the summary menu to analysis

The reason is that summarizing the workspace is more like an analysis, so it is not suitable to be placed in the File menu.

* Add an option for summary when opening the workspace. 

The reason is that in some cases, when there are a large number of classes inside, opening a workspace for analysis may result in lag or may not be necessary. Therefore, provide an option to choose whether to automatically generate a summary when opening the workspace.

* Adjust some inappropriate translations for entry-points.


Col-E:
Fixed the grammar mistakes, but otherwise 👍

PR #946, opened 2025-06-29
Add toggle for assembly/decompiled view in deobfuscation preview

luiox:
## What's new

* Add toggle for assembly/decompiled view in deobfuscation preview


luiox:
screenshot

image

PR #1025, opened 2026-05-27
Center subwindows on main window screen

SooStrator1136:
## What's new

Subwindows now center on the same screen as the main window when first shown

## What's fixed
Nothing


Col-E:
Couple of immediate concerns:

1. If you have a tall window like `System Information` - can this open the window such that the drag-bar is off-screen?
2. The `lastStageScreen` isn't updated to remove for non-anonymous windows that hide, and this includes any docking tab you 'pop-off' so anything externalized will stay in memory forever.

SooStrator1136:
>Couple of immediate concerns:
>
>    If you have a tall window like System Information - can this open the window such that the drag-bar is off-screen?
>    The lastStageScreen isn't updated to remove for non-anonymous windows that hide, and this includes any docking tab > you 'pop-off' so anything externalized will stay in memory forever.


I see you already addressed the points (and more) is there still something left to be done?

Col-E:
Nope, I have it merged locally - Just cleaning a few other things up first 👍

PR #1027, opened 2026-06-05
Fix keyboard shortcuts under non-Latin keyboard layouts

b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0:
## What's new

* Layout-independent keyboard shortcut handling for Ctrl/Meta + letter combinations
* Centralized key event normalization for text input controls and RichTextFX editors

## What's fixed

* Copy, paste, cut, and select-all not working in editors when a non-Latin keyboard layout was active
* Recaf keybindings (e.g. save, find, undo) failing to trigger under non-Latin layouts
* Paste-to-add-files shortcut in workspace builder not working with non-Latin layouts


Col-E:
> Layout-independent keyboard shortcut handling for Ctrl/Meta + letter combinations

https://github.com/FXMisc/RichTextFX/pull/1255 - IIRC this was fixed upstream for RichTextFX text area components _(Save, find, etc in the decompiled code editor)_

When we reported the issue, those combinations did work with alternative key layouts _(core JavaFX controls)_/ These combinations not working anymore in the places you've made fixes to seems to imply a regression upstream in core FX. Could you try and re-create the problem with these keys not working on a basic [hello-world](https://docs.oracle.com/javafx/2/get_started/hello_world.htm) scale JavaFX screen?

> Centralized key event normalization for text input controls and RichTextFX editors

I have no idea what this means.

b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0:
> Could you try and re-create the problem with these keys not working on a basic [hello-world](https://docs.oracle.com/javafx/2/get_started/hello_world.htm) scale JavaFX screen?

Tested with a minimal JavaFX repro (TextField, TextArea, RichTextFX CodeArea) on a non-Latin keyboard layout — Ctrl+C/V works in all three. So this doesn't look like a core JavaFX regression on my side; the issue appears Recaf-specific. I'll narrow the PR to Recaf keybinding matching and remove the scene-level event normalization.

b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0:
https://github.com/user-attachments/assets/5c72c387-345a-49c3-a358-a54e166fe93f

b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0:
> I have no idea what this means.

That line meant a scene-level filter that re-dispatched key events when Ctrl+letter produced a layout-specific KeyCode instead of the expected one. I've removed it. Hello-world testing showed copy/paste already works in core JavaFX and RichTextFX, so the PR now only covers Recaf keybinding matching.

Also added the missing Russian translations in the same PR. I know it would've been cleaner as a separate PR, but I already had this one open, so I just whipped up a quick Python script to sync the keys with en_US and filled in the translations.

recaf-ui/src/main/resources/translations/ru_RU.lang
```diff
@@ -195,15 +272,23 @@ dialog.filefilter.workspace=Рабочие пространства
...
 dialog.title.close-workspace=Закрыть рабочее пространство?
+dialog.title.promote-resource-workspace=Сделать основным ресурсом?
 dialog.option.create-workspace=Создать новое рабочее пространство
 dialog.option.update-workspace=Добавить в рабочее пространство
+dialog.header.promote-resource-workspace=Создать новое рабочее пространство с этим ресурсом как основным?
+dialog.content.promote-resource-workspace=Текущее рабочее пространство будет закрыто, а выбранный ресурс станет новым основным.
+
+Все открытые вкладки будут закрыты.
+
+Сохранённые изменения в рабочем пространстве останутся.
```
Col-E: Did you say these were automated? Some of these span multiple lines which isn't valid.

b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0: > Did you say these were automated?

Only the key matching was automate

b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0:
///

PR #1032, opened 2026-06-18
fix(analysis): fix inverted fabric mod initialization discovery logic

ghost:
## Description
Fixes a logical flaw and a flag-state bug in `FabricModEntryPointDiscovery` during the classpath scanning phase, which prevented accurate identification of Fabric mod entry points.

### Defect
The discovery mechanism suffered from an inverted state evaluation and an inconsistent loop short-circuit:
1. The `isInitializer(graph, className)` helper method incorrectly negated its assignment checks (`!graph.isAssignableFrom(...)`). As a result, it evaluated to `true` for arbitrary non-mod classes and `false` for genuine Fabric initializers.
2. The loop logic within `findEntryPoints` relied on `if (isModInitializer == null && (isModInitializer = isInitializer(...))) continue;`. Because of the inverted helper, a non-mod class with a single matching method name would trip the `continue` and avoid registration (accidentally masking the bug). However, if a non-mod class contained overloaded `onInitialize` methods, the second iteration skipped the `null` check, bypassed the `continue`, and registered a false-positive entry point.

### Fix
1. Corrected the boolean logic in `isInitializer` by removing the erroneous negation operators (`!`) and switching the conditional join from an AND (`&&`) to an OR (`||`).
2. Separated the lazy-initialization block from the control-flow filter inside the method loop. The analyzer now correctly evaluates the class type once, and explicitly issues a `continue` to skip the member methods only if `isModInitializer` is genuinely `false`.

### DIFF
--- a/recaf-core/src/main/java/software/coley/recaf/services/analysis/entry/FabricModEntryPointDiscovery.java
+++ b/recaf-core/src/main/java/software/coley/recaf/services/analysis/entry/FabricModEntryPointDiscovery.java
@@ -50,8 +50,12 @@ public class FabricModEntryPointDiscovery implements EntryPointDiscovery {
 				if (!methodName.equals("onInitialize") && !methodName.equals("onInitializeClient"))
 					continue;
 
-				// Check if this class is a mod-initializer subtype.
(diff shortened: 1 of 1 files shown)


Col-E:
Whoops!

PR #1044, opened 2026-08-12
Support larger (2GB+) files

Ian321:
## What's new

This PR replaces most `byte[]`  with `MemorySegment`  in order to support larger files.


Janmm14:
I'm not sure whether its easier to just use memorysegment directly and no longbytearray class

Col-E:
While the use of `MemorySegment` is nicer than the originally proposed `List<byte[]>` approach I still don't feel like this set of changes is worth the added complexity compared to the marginal edge cases it facilitates.

